PR #3, opened 2020-02-20
Initial, standalone service frontend with basic namespace calling out…

MikeCJohnsonUk:
… to stub handlers

This is an initial cut of a basic front end to show how this might be done. None of this is considered final and much is needed in the way of enhancement (e.g. user auth, sessions, callouts to the guts of the service, logging, error handling etc) but there should be enough to indicate how this frontend might be done.

There is also a lot of support code for the next few steps I plan to make so don't panic about code which is not apparently used.

As always feel free to comment.

### DIFF
--- a/internal/services/frontend/files.go
+++ b/internal/services/frontend/files.go
@@ -51,14 +51,14 @@ func serveFileErrorPage(code int, dir string) http.Handler {
 }
 
 func serveFileStatic(path string) http.Handler {
-	fp := filepath.Join(rootFilePath, path)
+	fp := filepath.Join(*rootFilePath, path)
 
 	return http.FileServer(http.Dir(fp))
 }
 
 func serveFileTemplate(w http.ResponseWriter, r *http.Request) {
-	lp := filepath.Join(rootFilePath, "templates", "layout.html")
-	fp := filepath.Join(rootFilePath, "templates", filepath.Clean(r.URL.Path))
+	lp := filepath.Join(*rootFilePath, "templates", "layout.html")
+	fp := filepath.Join(*rootFilePath, "templates", filepath.Clean(r.URL.Path))
 
 	// Return a 404 if the template doesn't exist
 	info, err := os.Stat(fp)
--- a/internal/services/frontend/files.go
+++ b/internal/services/frontend/files.go
@@ -1,0 +1,91 @@
+// This module containes the routines etc to implement the frontend handlers for the files
+// part of the API
+//
+
+//package frontend
+package main
+
+import (
+	"fmt"
+	"html/template"
+	"log"
+	"net/http"
+	"os"
+	"path/filepath"
+
+	"github.com/gorilla/mux"
+)
+
+func filesAddRoutes(routeBase *mux.Router) {
+
+	routeFiles := routeBase.PathPrefix("/static").Subrouter()
+
+	routeBase.HandleFunc("", handlerFilesRoot).Methods("GET")
+	routeBase.HandleFunc("/", handlerFilesRoot).Methods("GET")
+
+	routeFiles.HandleFunc("", handlerFilesRoot).Methods("GET")
+	routeFiles.HandleFunc("/", handlerFilesRoot).Methods("GET")
+}
+
+func handlerFilesRoot(w http.ResponseWriter, r *http.Request) {
+
+	fmt.Fprintf(w, "Files (Root)")
+	//	serveFileTemplate(w, r)
+
+	return
+}
+
+func serveFileErrorPage(code int, dir string) http.Handler {
+	fn := func(w http.ResponseWriter, r *http.Request) {
+
+		// Call duplicated by src/net/http/fs.go#L254
+		w.WriteHeader(code)
+
+		path := filepath.Join(dir, r.URL.Path)
+		log.Println(path)
+
+		http.ServeFile(w, r, path)
+	}
+
+	return http.HandlerFunc(fn)
+}
+
+func serveFileStatic(path string) http.Handler {
+	fp := filepath.Join(rootFilePath, path)
+
+	return http.FileServer(http.Dir(fp))
+}
+
+func serveFileTemplate(w http.ResponseWriter, r *http.Request) {
+	lp := filepath.Join(rootFilePath, "templates", "layout.html")
+	fp := filepath.Join(rootFilePath, "templates", filepath.Clean(r.URL.Path))
+
+	// Return a 404 if the template doesn't exist
+	info, err := os.Stat(fp)
+	if err != nil {
+		if os.IsNotExist(err) {
+			http.NotFound(w, r)
+			return
+		}
+	}
+
+	// Return a 404 if the request is for a directory
+	if info.IsDir() {
+		http.NotFound(w, r)
+		return
+	}
+
+	tmpl, err := template.ParseFiles(lp, fp)
+	if err != nil {
+		// Log the detailed error
+		log.Println(err.Error())
+		// Return a generic "Internal Server Error" message
+		http.Error(w, http.StatusText(500), 500)
+		return
+	}
+
+	if err := tmpl.ExecuteTemplate(w, "layout", nil); err != nil {
+		log.Println(err.Error())
+		http.Error(w, http.StatusText(500), 500)
+	}
+}
--- a/internal/services/frontend/frontend.go
+++ b/internal/services/frontend/frontend.go
@@ -1,0 +1,120 @@
+// This module contains the implementation of the service front end which establishes the http[s] endpoint to receive and handle incomging requests.
+//
+
+// Useful references
+//
+// https://golang.org/doc/articles/wiki/
+// https://tutorialedge.net/golang/creating-simple-web-server-with-golang/
+// https://gist.github.com/enricofoltran/10b4a980cd07cb02836f70a4ab3e72d7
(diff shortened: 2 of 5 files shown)


internal/services/frontend/frontend.go
```diff
@@ -0,0 +1,120 @@
...
+
+// StartService is the primary entry point to start the front-end web service.
+func StartService() error {
+
+	// Really need some error handling here since continuation from this
+	// point in the face of failure represents a compromise in security
+	//
+	if nil == keyAuthentication || nil == keyEncryption {
+		return ErrNotInitialized
+	}
```
Jim3Things: Shouldn't this just be a fatal error?  If I read the definition in the var section correctly, this cannot happen?  So wouldn't it be best as a fatalF() or equivalent call here?

internal/services/frontend/frontend.go
```diff
@@ -0,0 +1,120 @@
...
+	ErrUserPermissionDenied = errors.New("CloudChamber: permission denied")
+
+	dbComputers DbComputers
+
+	rootFilePath = "ROOT_PATH"
+)
+
+// StartService is the primary entry point to start the front-end web service.
+func StartService() error {
+
```
Jim3Things: Have you run these files through 'go fmt'?  I can't remember if it has an opinion about starting blank lines...

And 'go fmt' has issues that I'm struggling with in that it seems to generate a lot of file churn with no visible change... But at least that doesn't apply on new files ...

internal/services/frontend/frontend.go
```diff
@@ -0,0 +1,120 @@
...
+	//
+	filesAddRoutes(routeBase)
+	usersAddRoutes(routeAPI)
+	workloadsAddRoutes(routeAPI)
+
+	routeAPI.HandleFunc("/logs", handlerLogsRoot).Methods("GET")
+	routeAPI.HandleFunc("/stepper", handlerStepperRoot).Methods("GET")
+	routeAPI.HandleFunc("/injector", handlerInjectorRoot).Methods("GET")
+
+	http.ListenAndServe(":8080", routeBase)
```
Jim3Things: TODO: make the port a command line argument.

internal/services/frontend/frontend.go
```diff
@@ -0,0 +1,120 @@
...
+type Computer struct {
+	Name       string
+	Processors uint32
+	Memory     uint64
+}
+
+type DbComputers struct {
+	Lock      sync.Mutex
+	Computers map[string]Computer
+}
```
Jim3Things: Presumably these will be replaced in the future with etcd references to data stored using the protobuf definition?  I'm commenting mostly to assume that there's a TODO here...

internal/services/frontend/users.go
```diff
@@ -0,0 +1,215 @@
...
+	// evaluation. At somepoint, it will need to be removed, but in the meantime, leaving
+	// it there allows simple experimentation with just a browser.
+	//
+	routeUsers.HandleFunc("/{username}/add", handlerUsersAdd).Methods("PUT", "GET")
+	routeUsers.HandleFunc("/{username}/remove", handlerUsersRemove).Methods("DELETE", "GET")
+	routeUsers.HandleFunc("/{username}/enable", handlerUsersEnable).Methods("PUT", "GET")
+	routeUsers.HandleFunc("/{username}/disable", handlerUsersDisable).Methods("PUT", "GET")
+	routeUsers.HandleFunc("/{username}/update", handlerUsersUpdate).Methods("PUT", "GET")
+	routeUsers.HandleFunc("/{username}/login", handlerUsersLogin).Methods("PUT", "GET")
+	routeUsers.HandleFunc("/{username}/logout", handlerUsersLogout).Methods("PUT", "GET")
```
Jim3Things: I think I'd expect '/{username} with POST for add, PUT for update (or vice-versa, never can remember off the top of my head), and DELETE for remove, rather than further leaf addresses.  I'd also make enable, disable, login, and logout parameters rather than paths.

So, "POST .../Fred" would add Fred ,"DELETE .../Fred" would remove Fred, "PUT .../Fred?op=login" would log Fred in, etc.

internal/services/frontend/workloads.go
```diff
@@ -0,0 +1,127 @@
...
+
+	// In the following, the "GET" method is allowed just for the purposes of test and
+	// evaluation. At somepoint, it will need to be removed, but in the meantime, leaving
+	// it there allows simple experimentation with just a browser.
+	//
+	routeWorkloads.HandleFunc("/{workloadname}/add", handlerWorkloadsAdd).Methods("PUT", "GET")
+	routeWorkloads.HandleFunc("/{workloadname}/remove", handlerWorkloadsRemove).Methods("DELETE", "GET")
+	routeWorkloads.HandleFunc("/{workloadname}/enable", handlerWorkloadsEnable).Methods("PUT", "GET")
+	routeWorkloads.HandleFunc("/{workloadname}/disable", handlerWorkloadsDisable).Methods("PUT", "GET")
+	routeWorkloads.HandleFunc("/{workloadname}/update", handlerWorkloadsUpdate).Methods("PUT", "GET")
```
Jim3Things: Same comment as above

internal/services/frontend/workloads.go
```diff
@@ -0,0 +1,127 @@
...
+	routeWorkloads.HandleFunc("/{workloadname}/step", handlerWorkloadsSingleStep).Methods("PUT", "GET")
+	routeWorkloads.HandleFunc("/{workloadname}/step/{count}", handlerWorkloadsStep).Methods("PUT", "GET")
```
Jim3Things: I've viewed the stepper as a global function, not tied to a workload.  Not sure what it means to step a workload forward in time, but hold everything else...

internal/services/frontend/users.go
```diff
@@ -0,0 +1,215 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	user := User{Name: name, PasswordHash: passwordHash}
+
+	return &user, nil
+}
+
+func userAdd(name string, password []byte) error {
```
Jim3Things: As you complete this:  In the add path I think it would be good if you can spot that we're adding in the identical data to what is already there and return an alternate success back to the client.  Just thinking of ways for client-side retry to be simplified.

internal/services/frontend/files.go
```diff
@@ -0,0 +1,91 @@
...
+//
+
+//package frontend
+package main
+
+import (
+	"fmt"
+	"html/template"
+	"log"
+	"net/http"
```
Jim3Things: foo

internal/services/frontend/files.go
```diff
@@ -0,0 +1,91 @@
...
+
+	"github.com/gorilla/mux"
+)
+
+func filesAddRoutes(routeBase *mux.Router) {
+
+	routeFiles := routeBase.PathPrefix("/static").Subrouter()
+
+	routeBase.HandleFunc("", handlerFilesRoot).Methods("GET")
+	routeBase.HandleFunc("/", handlerFilesRoot).Methods("GET")
```
manpreet85: This all looks good.

internal/services/frontend/frontend.go
MikeCJohnsonUk: Agreed. Fixed.

internal/services/frontend/frontend.go
MikeCJohnsonUk: Agreed.
TODO added

internal/services/frontend/frontend.go
MikeCJohnsonUk: Exactly so.
TODO added.

internal/services/frontend/users.go
MikeCJohnsonUk: Makes sense.
I've updated to follow the CRUD template, but the parameters (for login etc. ) will have to wait for the next checkin as I need to understand the syntax in the gorilla package.

internal/services/frontend/workloads.go
MikeCJohnsonUk: As above. Updated to follow CRUD with parameter based commands to follow...

internal/services/frontend/workloads.go
MikeCJohnsonUk: The only real reason I'd even added this was to see if I properly understood the way the package parsed the path. I've remove this path.

internal/services/frontend/users.go
MikeCJohnsonUk: Is there such a thing as an alternate success in HTTP? Or does it have reason codes, extended status codes perhaps?

However, shouldn't we be careful to avoid adding a new user when there is already a user with the same name already present? Once we are really doing POST this should be an straight error right?

MikeCJohnsonUk: Further, I presume you mean completely identical, ie both the username and the supplied document containing the password etc, right? If so, then, with caution, I thinks that's ok, but I'm not entirely comfortable. I can see the point, but given that CREATE == POST which by defn is non-idempotent, I'm, well, I'm not sure where I sit.

Jim3Things: I'm not sure how to create the http status codes in the gorilla server, and I was thinking of 3 different cases:
- POST with no existing data: HTTP 201 (Created)
- POST with exact duplicate data: HTTP 200 (OK)
- POST with different data: an HTTP error code, probably 409 (Conflict)

Not sure what that would translate into for Gorilla.  Btw, I do fully agree that the duplicate case should not return the same value as the new-create case.  Fyi.

PR #5, opened 2020-02-27
Create fetch_imports.cmd

Jim3Things:
I've added the import commands to set up the Go dependencies.

(command file almost completely based on Mike's work)


manpreet85:
I have copied and pasted all the go commands in Gitcmd and all the commands got executed. But I have not tested the modules explicitly. But if the test was part of the file than I would need your guidance to understand the output.

manpreet85:
Looks good to me as well.

PR #7, opened 2020-02-29
Add timestamp unit test, minor protobuf cleanup

Jim3Things:
Some minor cleanup tasks


internal/clients/timestamp.go
```diff
@@ -4,16 +4,17 @@
 package clients
 
 import (
-    "context"
-    "time"
+	"context"
+	"time"
```
MikeCJohnsonUk: This time the spacing seems to have gone from 4 space chars to a single tab?

internal/clients/timestamp.go
```diff
@@ -22,77 +23,109 @@ var dialOpts []grpc.DialOption
...
+	if err != nil {
+		return err
+	}
+
+	defer func() { _ = conn.Close() }()
+
+	client := pb.NewStepperClient(conn)
+
+	_, err = client.SetPolicy(ctx, &pb.PolicyRequest{Policy: policy, MeasuredDelay: delay})
+
```
MikeCJohnsonUk: Just checking, ... so if I understand things correctly, after establishing a connection from this (the test client) to the service, this is a call to that service using a protobuf message built by the PolicyRequest() call right?

internal/clients/timestamp_test.go
```diff
@@ -1,23 +1,111 @@
...
+	"log"
+	"net"
 	"testing"
+	"time"
+
+	"github.com/golang/protobuf/ptypes/duration"
+	"github.com/stretchr/testify/assert"
+
+	"github.com/Jim3Things/CloudChamber/internal/services/stepper"
+	ctrc "github.com/Jim3Things/CloudChamber/internal/tracing/client"
```
MikeCJohnsonUk: ctrc ?

internal/clients/timestamp_test.go
```diff
@@ -1,23 +1,111 @@
...
+	"github.com/Jim3Things/CloudChamber/internal/tracing/exporters"
+	"github.com/Jim3Things/CloudChamber/internal/tracing/exporters/unit_test"
+	strc "github.com/Jim3Things/CloudChamber/internal/tracing/server"
+	"github.com/Jim3Things/CloudChamber/internal/tracing/setup"
+	pb "github.com/Jim3Things/CloudChamber/pkg/protos/Stepper"
+	ct "github.com/Jim3Things/CloudChamber/pkg/protos/common"
```
MikeCJohnsonUk: ct ?

internal/clients/timestamp_test.go
```diff
@@ -1,23 +1,111 @@
...
+	"time"
+
+	"github.com/golang/protobuf/ptypes/duration"
+	"github.com/stretchr/testify/assert"
+
+	"github.com/Jim3Things/CloudChamber/internal/services/stepper"
+	ctrc "github.com/Jim3Things/CloudChamber/internal/tracing/client"
+	"github.com/Jim3Things/CloudChamber/internal/tracing/exporters"
+	"github.com/Jim3Things/CloudChamber/internal/tracing/exporters/unit_test"
+	strc "github.com/Jim3Things/CloudChamber/internal/tracing/server"
```
MikeCJohnsonUk: Ah, so perhaps ctrc and strc are client and server trace respectively?

internal/clients/timestamp_test.go
```diff
@@ -1,23 +1,111 @@
...
+
+		data := <-ch
+		afterHit = true
+
+		assert.Nil(t, data.err)
+		assert.GreaterOrEqual(t, deadline, data.time.Ticks)
+	}(3)
+
+	assert.Nil(t, Advance())
+	time.Sleep(time.Duration(2) * time.Second)
```
MikeCJohnsonUk: Just confirming, this sleep is just to let the system achieve steady state right?

pkg/protos/workload/external.proto
```diff
@@ -5,19 +5,14 @@ package workload;
...
 
 message external {
-
-    // TODO: Should we make this a common type?
-    message feature {
-        string name = 1;
-        int64 amount = 2;
-    }
+    // TODO: Open question.  Given the features of protobuf, I'm not sure that a generic 'features' collection
+    // makes sense.  I've left it in here for now, but I think we will probably remove it.
```
MikeCJohnsonUk: I think I agree. "feature" would seem to be related to the underlying type of the thing and although different things might have "features" which at least superficially look the same, there is no reason to believe that are the same, or behave in the same way etc. Seems unfortunate perhaps, but there it is.

internal/clients/timestamp.go
Jim3Things: Weirdly, it didn't in my IDE.  I frankly can't make sense of the line ending & indention changes that seem to happen depending on tooling...

internal/clients/timestamp.go
Jim3Things: Yes.

internal/clients/timestamp_test.go
Jim3Things: client trace.  Just trying to follow Go's recommendations for short names.  Don't like the name, but also didn't like client_trace or any equivalent... Open to alternatives.

internal/clients/timestamp_test.go
Jim3Things: common types

internal/clients/timestamp_test.go
Jim3Things: Yes.

internal/clients/timestamp_test.go
Jim3Things: Yes.

pkg/protos/workload/external.proto
Jim3Things: Yeah.  I think that given Protobuf's support for optional fields we'll end up just adding fields in the message, one for each unique 'feature'.  Rather than a key-value collection of feature names & values...

PR #9, opened 2020-03-11
Add protoc-gen-validate support, protobuf cleanup

Jim3Things:
Added support for the protoc-gen-validate annotations.
Updated stepper.go to use these validations.
Updated the proto files with further elaboration on the models.  Commenting.


pkg/protos/common/capacity.proto
```diff
@@ -21,19 +23,34 @@ message blade_capacity {
...
-    // GPUs may not be present at all on a blade, so if they are
-    // not then the 'present' flag is false and the other fields are
-    // ignored.
-    message gpu_details {
-        int32 units = 1;
-        string arch = 2;
-        bool present = 3;
+    // The number of cores on the blade.  This is defined as a float to allow easier comparison to workload
+    // requirements that specify partial cores (e.g. through time sliced sharing)
+    float cores = 1 [(validate.rules).float.gt = 0];
```
MikeCJohnsonUk: Whilst I can see why you might want to be able to specify the workload in terms of fractional cores, I'm not sure I agree that the definition of a blade should allow the same. Given that if this is defined as a float, it will have to be verified later on to ensure it's not a non-integral number, why not just define as and int and convert to a float internally?

pkg/protos/common/capacity.proto
```diff
@@ -21,19 +23,34 @@ message blade_capacity {
...
+
+    // The amount of memory, in megabytes
+    int64 memory_in_mb = 2 [(validate.rules).int64.gt = 0];
+
+    // The amount of local disk space, in gigabytes.  Note that this assumes either one disk, or that the disks are
+    // mounted collectively as a single volume
+    int64 disk_in_gb = 3 [(validate.rules).int64.gt = 0];
```
MikeCJohnsonUk: I have to admin I'm torn on this. I mean a GB seems like such a large unit, but does anyone in the real world allocate in terms of anything much smaller than this?

pkg/protos/common/capacity.proto
```diff
@@ -21,19 +23,34 @@ message blade_capacity {
...
+
+    // The network bandwidth from the host in megabytes per second
```
MikeCJohnsonUk: MB or Mb? b/w is normally described in Mbps, Gbps and presumably soon we'll also have Tbps.

pkg/protos/inventory/external.proto
```diff
@@ -25,29 +29,20 @@ message external {
...
+        // specify the blades in the rack.  Each blade is defined by an integer index within that rack, which is used
+        // here as the key.
+        map<int64, common.blade_capacity> blades = 3 [(validate.rules).map.min_pairs = 1];
     }
 
-    // This is probably not needed at this point.
+    // Finally, a zone is a collection of racks.  Each rack has a name, which is used as a key in the map below.
     message zone {
-        repeated rack racks = 1;
+        map<string, rack> racks = 1 [(validate.rules).map.min_pairs = 1];
```
MikeCJohnsonUk: So how does storage show up? I presume the mention above was for direct-attached devices to a blade. What about a storage shelf?

By the way, do we group blades into shelves or just associate them with the rack? If just with the rack, do/can/should we do the same for storage devices?

pkg/protos/workload/external.proto
```diff
@@ -2,37 +2,81 @@ syntax = "proto3";
...
-        int64 memory_in_mb = 2;
+            // Volume exists, and is writable
+            Writable = 3;
+        }
+
+        // Volume size in gigabytes
+        int64 size_in_gb = 1 [(validate.rules).int64.gt = 0];
+
+        // Constraint on the volume type.
+        attr attribute = 2 [(validate.rules).enum.defined_only = true];
```
MikeCJohnsonUk: Are we modelling in terms of just volumes, or down to the storage device level?

pkg/protos/common/capacity.proto
```diff
@@ -21,19 +23,34 @@ message blade_capacity {
...
+
+    // Define the set of known accelerators, such as GPUs or FPGAs.
+    message accelerator {
+        message NVIDIA_V100 { }
+
+        oneof accelerator_type {
+            option (validate.required) = true;
+
+            NVIDIA_V100 v100 = 1;
+        }
```
MikeCJohnsonUk: I presume we're intending to add a whole bunch of other accelerators in here?

pkg/protos/common/capacity.proto
Jim3Things: You're probably right.  I was trying to leverage the same structure in two places because they're so similar structurally, but they actually do mean different things.  I'll split the uses apart and make this one an int64.

pkg/protos/common/capacity.proto
Jim3Things: yeah, I couldn't see the reason to allow for smaller.  We can always change it later if we need to.

pkg/protos/common/capacity.proto
Jim3Things: Good catch - should be Mb.

pkg/protos/inventory/external.proto
Jim3Things: The model is currently incomplete, but I expect to model storage as an external service.  Thinking about Azure's offering, there is local disk on a blade and 'disks' or 'volumes' that come from the storage service.  That was my thought here - that the storage volumes used by a workload came from a service that was essentially off-stage.  Does that make sense?

pkg/protos/workload/external.proto
Jim3Things: See above -- current thought is that workloads model volumes, which are present via an external storage service.

pkg/protos/common/capacity.proto
Jim3Things: Yup.  This is a pro-forma placeholder to show the structure.

pkg/protos/common/capacity.proto
Jim3Things: Should be fixed in the latest rev.

pkg/protos/inventory/external.proto
Jim3Things: I've added URI for where to get the volumes, but there are a lot of details still missing (e.g. security).

pkg/protos/common/capacity.proto
MikeCJohnsonUk: Fair enough.

pkg/protos/inventory/external.proto
MikeCJohnsonUk: Ah, I see. Ok. Well, that does simplify the storage aspect somewhat as there are way too many ways to construct volumes from underlying devices.

pkg/protos/workload/external.proto
MikeCJohnsonUk: Got it.

PR #11, opened 2020-03-18
New tools and updates to fetch_imports.cmd

MikeCJohnsonUk:
Some tools updates:
Add .gitignore to let git know about Visual Studio Code metadata files
Add UpdateAll.cmd to update all the imports
Add StartEtcd.cmd to start a single ETCD instance from locally built sources.
Add FetchAll.cmd - basically a copy of fetch_imports.cmd at present. Long term will supercede fetch_imports.cmd
Update fetch_imports.cmd

I was having trouble fetching/updating various packages so made some updates. Works for me but I'd appreciate if folks could check it works for them.

NOTE to Jim - there's a reference you added to "make" to build the protobuf validator but that's not a standard (Windows) tool so it currently errors out for me. I can probably find one as part of the GNU toolset but I was wondering where yours came from.


tools/fetchall.cmd
```diff
@@ -0,0 +1,37 @@
...
+go get github.com/stretchr/testify/assert
+
+go get go.opentelemetry.io/otel
+
+go get google.golang.org/grpc
+
+go get golang.org/x/crypto/...
+
+
+go get githbu.com/Jim3Things/CloudChamber
```
Jim3Things: typo for sure here.  githbu.com...

Jim3Things:
Mike,

I realize that I just used the terminal window in GoLand.  I don't know
where it pulled its make from.  I do have Cygwin installed, but I don't
think that would have come into play here.

Looking at the makefile, I see:

Jim3Things:
.. continuing.  Looking at the make file I see:




.PHONY: build

build: validate/validate.pb.go
# generates the PGV binary and installs it into $$GOPATH/bin
go install .


So maybe we just need 'go install .'  Not sure what to make of the
validate/validate.pb.go, unless that is the build trigger...

Thoughts?
Jim.

tools/updateall.cmd
```diff
@@ -0,0 +1,29 @@
...
+
+go get -u github.com/stretchr/testify/assert
+
+go get -u go.opentelemetry.io/otel
+
+go get -u google.golang.org/grpc
+
+go get -u golang.org/x/crypto/...
+
+popd
```
Jim3Things: Looking at the 'go get' documentation ('go help get'), it doesn't look like -u is quite what I'd expect.  The help text says:

The -u flag instructs get to use the network to update the named packages
and their dependencies. By default, get uses the network to check out
missing packages but does not use it to look for updates to existing packages.

Since I've successfully used -u while updating the opentelemetry packages, I'm not sure what to make of this.

tools/fetchall.cmd
MikeCJohnsonUk: Ah, ..... oops? Unless githbu really is a valid site ;-)

Fixed.

MikeCJohnsonUk:
I tried the earlier cmd files because in attempting to resolve my NFS/Windows User creds mapping I ended up deleting large paths of the local store and re-fetching them. 

Or not when the fetches failed.

Which is what lead to my tinkering.

In terms of the make file, if I read it correctly we would need to build validate/validate.pb.go first (go build ?) and then the "go install ." right?

Mike

Jim3Things:
If I look at the code in the GitHub repository, validate.pb.go is already
built, so I'm not sure.

Jim.

Jim3Things:
I'm ok with the other changes, not sure yet what to do with the make command.  Let's discuss, but I'm ok with pre-emptively merging the rest.

MikeCJohnsonUk:
Ok, that works for me too. I've really been treating the CMD tools as partly reference material, partly examples for the present. So not something I'm going to get too excited about. 

I'll complete the PR.

PR #13, opened 2020-03-27
Global configuration file parsing

Jim3Things:


internal/config/settings.go
```diff
@@ -0,0 +1,186 @@
...
+        "ReadGlobalConfig")
+    defer span.End()
+
+    if err := viper.ReadInConfig(); err != nil {
+        if _, ok := err.(viper.ConfigFileNotFoundError); ok {
+            // Config file not found; we'll just use the default values
+            span.AddEvent(
+                ctx,
+                fmt.Sprintf(
+                    "No config file found at %s/%s (%s), applying defaults.\n",
```
MikeCJohnsonUk: Probably minor, but should trace message strings themselves have the "\n"?

tools/fetch_imports.cmd
```diff
@@ -34,4 +34,6 @@ go get google.golang.org/grpc
 
 go get golang.org/x/crypto/...
 
+go get github.com/spf13/viper
```
MikeCJohnsonUk: Are we at the point we can remove fetch_imports.cmd in favour of fetchall.cmd? That was certainly my intent when adding fetchall.cmd

internal/config/settings.go
```diff
@@ -0,0 +1,186 @@
...
+        // Fill in the global configuration object from the configuration file
+        if err = viper.UnmarshalExact(cfg); err != nil {
+            err = fmt.Errorf("unable to decode into struct, %v", err)
+            span.AddEvent(ctx, err.Error())
+            return nil, err
+        }
+    }
+
+    span.AddEvent(ctx,
+        fmt.Sprintf("Configuration Read: \n%v", ToString(cfg)))
```
MikeCJohnsonUk: \n before the %v?

internal/config/settings.go
Jim3Things: Good catch.  Will fix.

tools/fetch_imports.cmd
Jim3Things: I think so.  Fwiw, I've been using updateall exclusively.

internal/config/settings.go
Jim3Things: That's because the ToString produces a formatted output, the \n ensures that the first line formatting is correct.

Jim3Things:
We can add a command line qualifier ('-showconfig') that would write out the effective config to stdout by using the ToString function.  I'll add that in the next update.

Jim3Things:
I've updated with the CR comments.

internal/config/settings.go
```diff
@@ -0,0 +1,186 @@
...
+    DefaultGlobalConfigFile = "cloudchamber.yaml"
+    DefaultConfigType = "yaml"
+    ControllerDefaultPort = 8081
+    InventoryDefaultPort = 8082
+    SimSupportDefaultPort = 8083
+    WebServerDefaultPort = 8084
+    WebServerFEDefaultPort = 8080
+    DefaultHost = ""
+    DefaultRootFilePath = "."
+    DefaultSystemAccount = "Admin"
```
MikeCJohnsonUk: Implies user/account names are not case-sensitive?

Jim3Things: No.  I don't think this component cares.  I've set up some defaults that are supplied if there's no config here, and the .yaml also has values provided.  Some of them are different specifically so I could verify that the .yaml was processed and that I wasn't looking at the result of the default pass.  I figured some of these would change as we get, e.g. the full user support in place.

PR #14, opened 2020-03-31
Initial version for store APIs

MikeCJohnsonUk:
This is the first version of the basic APIs to allow IO to a backing etcd store along with a bunch of test code which should be sufficient to demonstrate their use.

At present only basic ReadXxx() WriteXxx() and DeleteXxx() methods are included with more to come later. However, there should be enough here to allow some simple IO calls.

I'd be interested in comments as to language usage in addition to the normal review type observations.

Note: this change has also moved the timestamp client to one directory lower down to avoid package cross-contamination.


internal/clients/store/store.go
```diff
@@ -0,0 +1,657 @@
...
+//
+// This method can only be used when the store object is not connected.
+//
+func (store *Store) SetAddress(endpoints []string) (err error) {
+
+	if nil != store.Client {
+		err = ErrStoreConnected
+		log.Printf("Cannot set the endpoint if the store is currently connected - error: %v endpoints: %v", err, endpoints)
+		return err
+	}
```
Jim3Things: I'm not sure how much this will help, but it seems that you have this paragraph repeatedly throughout - can you make into a helper function?  e.g. if err = disconnected(store); err != nil { return err }

... and the inverse case for the correct other methods as well.

Yeah, I miss some of the Bliss macro conventions of the past...

internal/clients/store/store.go
```diff
@@ -0,0 +1,657 @@
...
+	cancel()
+
+	if err != nil {
+		logEtcdResponseError(err)
+		log.Printf("Unable to write all the key/value pairs - requested: %v achieved: %v", len(keyValueSet), processedCount)
+	}
+
+	for i := 0; i < processedCount; i++ {
+		log.Printf("wrote/updated [%v/%v] key: %v value: %v", i, processedCount, keyValueSet[i].key, keyValueSet[i].value)
+	}
```
Jim3Things: When we put lines like these into the trace path, assuming that we do, we'll want to make sure that they are at a debug or verbose level.  I think that we should also think about how to scrub for PII, but that is not needed just now.

internal/clients/store/store.go
```diff
@@ -0,0 +1,657 @@
...
+
+	log.Printf("Processed %v items", processedCount)
+
+	return err
+}
+
+// Read is a method to read a single value from the store using the supplied key.
+//
+// It is expected that the store will already have been initialized and connected
+// to the backed db server.
```
Jim3Things: backed -> backend?

internal/clients/store/store.go
```diff
@@ -0,0 +1,657 @@
...
+
+	if nil == store.Client {
+		err := ErrStoreNotConnected
+		log.Printf("Failed to write - no current connection - error: %v", err)
+		return err
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), store.TimeoutRequest)
+	_, err := store.Client.Put(ctx, key, value)
+	cancel()
```
Jim3Things: Are there forms of WithTimeout that take a function as an argument (e.g. like trace.WithSpan)?

internal/clients/store/store.go
```diff
@@ -0,0 +1,657 @@
...
+	if err != nil {
+		logEtcdResponseError(err)
+		log.Printf("Unable to read all the key/value pairs - requested: %v achieved: %v", len(keySet), processedCount)
+	} else {
+		results = make([]KeyValueResponse, processedCount)
+
+		for i := 0; i < processedCount; i++ {
+			results[i].key = string(responses[i].Kvs[0].Key)
+			results[i].value = responses[i].Kvs[0].Value
+		}
```
Jim3Things: I honestly don't know how etcd works sufficiently, so this is a question.  Can I constrain the read to only get Kvs[0] values?  Is it an issue if the code above gets more than one column of Kvs data?  Just not sure at this point.

internal/clients/store/store.go
```diff
@@ -0,0 +1,657 @@
...
+
+	ctx, cancel := context.WithTimeout(context.Background(), store.TimeoutRequest)
+	response, err := store.Client.Delete(ctx, key)
+	cancel()
+
+	if err != nil {
+		logEtcdResponseError(err)
+	} else if 1 != response.Deleted {
+		err = ErrStoreBadResultSize
+		log.Printf("expected a single deletion and instead received something else - error: %v expected: 1 received: %v\n", err, response.Deleted)
```
Jim3Things: This might be another good case for an error type that takes some parameters.

internal/clients/store/store_test.go
```diff
@@ -0,0 +1,604 @@
...
+
+	endpoints := defaultEndpoints
+	timeoutConnect := defaultTimeoutConnect
+	timeoutRequest := defaultTimeoutRequest
+
+	store, err := New(endpoints, timeoutConnect, timeoutRequest)
+
+	assert.Nilf(t, err, "Failed to allocate new store - error: %v", err)
+	assert.NotNilf(t, store, "Failed to get the store as expected - error: %v", err)
+
```
Jim3Things: This brings up that it would be good to have store.GetXxx methods and verify here (and throughout) that they are what they are expected to be..

internal/clients/store/store_test.go
```diff
@@ -0,0 +1,604 @@
+// Unit tests for the web service store package
+//
```
Jim3Things: I've got a longer term question.  Can the etcd itself be sufficiently mocked out so that at least some of these UTs can be run without running etcd?  I realize that there are points where we must use a real etcd - just wondering if there's a subset where we can avoid it.  And yes, a longer term question.

internal/clients/store/store.go
MikeCJohnsonUk: I'll do something about this.

MikeCJohnsonUk: Done

internal/clients/store/store_test.go
MikeCJohnsonUk: I've seen reference to a notion of an embedded etcd which strike me as interesting. I've had some thought about whether we might be able to make use of such a component with perhaps a config option to allow the use of an external etcd should the user choose. That should make things simpler for the more bare-bones install case.

Would that be useful? I think I'd prefer that than trying to substitute a replacement.

internal/clients/store/store.go
```diff
@@ -0,0 +1,657 @@
...
+	// ErrStoreBadResultSize indicates the size of the result set does not match
+	// expectations. There may be either too many, or too few. Typically a single
+	// result way anticipated and more that that was received.
+	//
+	ErrStoreBadResultSize = errors.New("CloudChamber: unexpected size for result set")
+
+	// ErrStoreNotImplemented indicated the called method does not yet have an
+	//implementation
+	//
+	ErrStoreNotImplemented = errors.New("CloudChamber: method not currently implemented")
```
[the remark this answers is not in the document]
MikeCJohnsonUk: Interesting. I see a few places where this would be useful.

internal/clients/store/store.go
```diff
@@ -0,0 +1,657 @@
...
+// New is a method supplied values. These values can later be overridden with
+//
+//    SetAddress()
+//    SetTimeoutConnection()
+//    SetTimeoutResponse()
+//
+// providing the store has not yet connect3d to the back-end db service.
+//
+func New(endpoints []string, timeoutConnect time.Duration, timeoutRequest time.Duration) (*Store, error) {
+
```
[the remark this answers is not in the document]
MikeCJohnsonUk: Hhhmmm, I'm a little more cautious. I've spent waaaayyy to much time not having a status to begin with and then having to go back and add one. 

For the simple cases such as here which are unlikely to get more complicated later, sure.

internal/clients/store/store.go
MikeCJohnsonUk: I'd already spotted the need for a "verbosity" level and it's on my todo list.

As for PII - how much care is needed here really? I presume you are thinking of cases where a User might submit a trace back to us for debug?

internal/clients/store/store_test.go
Jim3Things: That might be, yeah.  And if it would make initial config easier, so much the better.

internal/clients/store/store.go
Jim3Things: This is more about showing what you'd want to think about - the traces are things we'll produce, but some fields, such as account name, might be considered PII - so how would we best handle that?

internal/clients/store/store.go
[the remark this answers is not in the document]
Jim3Things: Yeah, point taken.

internal/clients/store/store.go
```diff
@@ -0,0 +1,657 @@
...
+	return &store, err
+}
+
+// Initialize is a method used to initialize a specific instance of a Store struct.
+//
+func (store *Store) Initialize(endpoints []string, timeoutConnect time.Duration, timeoutRequest time.Duration) (err error) {
+
+	if nil != store.Client {
+		err = ErrStoreConnected
+		log.Printf("Unable to initialize a store that is already in use - error: %v", err)
```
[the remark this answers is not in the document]
MikeCJohnsonUk: Agreed. I'll see if I can get this done for this PR, if not, I'll do a separate follow-on.

internal/clients/store/store.go
MikeCJohnsonUk: A number of questions come to mind.

1. Who is going to see the output?
2. Given a need to disambiguate an account (not a foregone conclusion) what might be considered an acceptable substitute? We could, for example, generate a dynamic userAccountTraceId which is unique to an execution of the service, not preserved anywhere etc. I'd not really planned on even needing a userId and instead just relying on a userName.
3. I've no idea what the public clouds do for this sort of thing?
4. What about things like workload names etc. How far would we need to consider taking this?
5. We would need start thinking about separating what is required for trace versus what needs to be logged for auditing etc.

internal/clients/store/store.go
MikeCJohnsonUk: Seems not to. Why do you ask?

internal/clients/store/store.go
MikeCJohnsonUk: For cases where a single request can result in multiple values in a response, yes, there is a means of constraint. I should probably add this for the ReadWithPrefix() and probably the DeleteWithPrefix() funcs.

The other calls (e.g. ReadMultiple()), are really a sequence of one or more calls each dealing with individual/single results. So in the case above, we have an array of responses with a single value in each response. That is, we make a set of individual calls which have a response containing a single value in each. We just accumulate all the responses once we are done with the individual calls.

Unlike the ReadWithPrefix() case where we have a single response with 0 or more values. The difference is triggered by the WithPrefix() option in the Get() calls. Another option is WithLimit() which allows setting a limit in the nuber of values in the response, and WithRange() is probably what is used to enumerate over limited size fetches when the matching value set is very large.

Also, there are no columns. Each key has a single []bytes slice value. It's all or nothing.

internal/clients/store/store.go
MikeCJohnsonUk: Agreed.

internal/clients/store/store_test.go
MikeCJohnsonUk: Not sure I understand the comment.

internal/clients/store/store.go
Jim3Things: Good questions, and probably this is a todo that we need to discuss in more detail.

The only reason I brought it up is for using the code to show what 'real' clusters would contend with.  In the case of a public service, protecting PII from accidental exposure is something one would worry about.  But it is a longer point and one that I'm happy to defer out of this change.  Let's talk about it at our next meetup.

internal/clients/store/store.go
Jim3Things: Just don't like the cancel() boilerplate, and was wondering if WithTimeout already encapsulated that.  If it doesn't, then this is fine.

internal/clients/store/store.go
Jim3Things: First, probably bad syntax on my part.  I was thinking of [0] as a column, probably the wrong word.  What I was asking has to do with this: when I see, e.g., Kvs[0] I wonder if there's a Kvs[1], and what it means if there is or isn't.  That was really the root of my question - would a Kvs[1] be expected, allowed, or a surprise.  If the ReadXxx can constrain the call to only have a Kvs[0], then that answers the question.  If not, then there's just my starting question.  HTH.

internal/clients/store/store_test.go
Jim3Things: Just to verify that e.g. New(...) sets the values the way we expect.

MikeCJohnsonUk: Oh, I see.

internal/clients/store/store.go
MikeCJohnsonUk: I didn't much like it either. Apparently it's there to cleanup the ctx and it's associated resources . Seems like an odd paradigm myself.

I mean you *could* defer it, but that then has the ctx (and associated resources) hanging around until function exit. I don't think it's any better.

Jim3Things: Yeah, I don't think that is much better either.

internal/clients/store/store.go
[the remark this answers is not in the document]
MikeCJohnsonUk: Done. I think....

internal/clients/store/store.go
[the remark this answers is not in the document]
MikeCJohnsonUk: Done the simple cases, left the others

internal/clients/store/store.go
MikeCJohnsonUk: I believe I understand. To help clarify

In the Read() case, we issue a single Get() for a single key/value pair. Either you get an empty response (handled by logEtcdResponseError()) or you get a single key/value pair which implies the Kvs[0] entry. For a correctly functioning etcd there MUST NOT be more than a single entry. We verify that with the check that leads to the ErrStoreBadResultSize error.

In the ReadWithPrefix() case (i.e. a wildcard fetch), we expect anything from 0 to n responses and so process them as such.

The ReadMultiple() is just a convenience wrapper around the single READ, but it seems I neglected the single result validation. I'll add it.


As for the result set size from ReadWithPrefix(), yes it can be constrained, but currently we don't do that. There is an inherent response size limit so a sufficiently large set could hit it. We will need to add something here. Looking at the available options on (say) the Get(), it's not as clean as one might like, i.e. there's no simple enumeration mechanism, at least not one that I can see. I do have a potential solution in mind, but it likely involved more queries to the etcd service than I'd like and I think we'll need to think about this further. Do we have a good feel for how many keys under a common prefix that one might reasonably have?

internal/clients/store/store.go
MikeCJohnsonUk: Done.

internal/clients/store/store.go
Jim3Things: Ok, thanks.

Don't have a great idea just now, but I'd expect 100's in inventory, 100's in workloads at least to start.  Maybe 10x later.  It is just a simulation after all :)

MikeCJohnsonUk: Fix for the ReadMultiple() case done.

internal/clients/store/store.go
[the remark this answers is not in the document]
MikeCJohnsonUk: Deferred to a subsequent check-in

internal/clients/store/store.go
MikeCJohnsonUk: Verbosity level added.

internal/clients/store/store_test.go
MikeCJohnsonUk: store.GetXxx() methods added along with appropriate tests.

internal/clients/store/store_test.go
MikeCJohnsonUk: Embedded etcd on the ToDo list.

PR #15, opened 2020-04-01
first part of milestone 1 E2E test

Jim3Things:
This is the initial path for the M1 E2E test.  It only works up through the login call.  Later changes will continue on, once we have the inventory calls in place.

I've updated frontend.go & users.go to fix some items, fill out the users implementation as needed, and some cleanup.  Please take a look and see if these make sense.

I've updated the frontend_test.go to account for changes in behavior, and to integrate better with the tracing system.  Again, I'd appreciate any comments on that.

Finally, there are some changed and additional .cmd files, especially one that now launches CC.

### DIFF
--- a/internal/services/frontend/frontend_test.go
+++ b/internal/services/frontend/frontend_test.go
@@ -14,6 +14,8 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+
+	"github.com/Jim3Things/CloudChamber/internal/config"
 )
 
 var (
@@ -30,7 +32,20 @@ func commonSetup() error {
 		log.Fatalf("Error initializing service for second or subsequent time")
 	}
 
-	if err := initService(); err != nil {
+	if err := initService(&config.GlobalConfig{
+		Controller: config.ControllerType{},
+		Inventory:  config.InventoryType{},
+		SimSupport: config.SimSupportType{},
+		WebServer:  config.WebServerType{
+			RootFilePath:  "C:\\CloudChamber",
+			SystemAccount: "Admin",
+			FE:            config.Endpoint{
+				Hostname: "localhost",
+				Port:     8080,
+			},
+			BE:            config.Endpoint{},
+		},
+	}); err != nil {
 		log.Fatalf("Error initializing service: %v", err)
 	}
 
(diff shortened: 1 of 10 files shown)


internal/services/frontend/frontend_test.go
```diff
@@ -24,21 +29,42 @@ var (
...
+	if err := initService(&config.GlobalConfig{
+		Controller: config.ControllerType{},
+		Inventory:  config.InventoryType{},
+		SimSupport: config.SimSupportType{},
+		WebServer:  config.WebServerType{
+			RootFilePath:  "C:\\CloudChamber",
+			SystemAccount: "Admin",
+			FE:            config.Endpoint{
+				Hostname: "localhost",
+				Port:     8080,
```
MikeCJohnsonUk, line 48: Something funky with the spacing? The hostname and port fields are indented deeper than the rest.

internal/services/frontend/users.go
```diff
@@ -119,54 +155,56 @@ func usersAddRoutes(routeBase *mux.Router) {
     //   PUT is idempotent so translates to UPDATE in the CRUD methodolgy
     //   POST is NOT idempotent so translates to CREATE in the CRUD methodolgy
     //
+    routeUsers.HandleFunc(routeString, handlerUsersOperation).Queries("op", "{op}").Methods("PUT")
```
MikeCJohnsonUk, line 158: I thought I didn't have this done properly. I like this better.

internal/services/frontend/users.go
```diff
@@ -119,54 +155,56 @@ func usersAddRoutes(routeBase *mux.Router) {
...
+        fmt.Fprintf(w, "User: %v", username)
     }
 }
 
+// Process an http request for the list of users.  Response should contain a document of links to the
+// details URI for each known user.
 func handlerUsersList(w http.ResponseWriter, r *http.Request) {
-
-    fmt.Fprintf(w, "Users (List)")
+    _ = tr.WithSpan(context.Background(), "HandleUsersUpdate", func(ctx context.Context) error {
```
MikeCJohnsonUk, line 180: Is the comment supposed to reflect the func name or something else? If the func name can we use Name() in the runtime package assuming there is no equivalent in the telemetry package?

test/Milestone1/TestClient/main.go
```diff
@@ -0,0 +1,120 @@
...
+    if *showConfig {
+        fmt.Println(config.ToString(cfg))
+        os.Exit(0)
+    }
+
+    baseAddress := fmt.Sprintf("http://%s:%d/api", cfg.WebServer.FE.Hostname, cfg.WebServer.FE.Port)
+    client := &http.Client{}
+
+    // 0: get list of known users
+    target := fmt.Sprintf("%s/users/admin", baseAddress)
```
MikeCJohnsonUk, line 52: Do we mean currently know users, or just the pre-defined one?

At some point I'm anticipating that we'll be adding users to a persistent store somewhere, so I would expect "known" to include all those already created.

MikeCJohnsonUk: Ok.

internal/services/frontend/users.go

internal/services/frontend/frontend_test.go
MikeCJohnsonUk: Ah. Ok.

PR #16, opened 2020-04-04
Msm users frontend.go

manpreet85:
Inventory.go based on workloads.go

It is skeletal and would need more work before it can be usefule


internal/services/frontend/doc.go
```diff
@@ -101,12 +101,37 @@ Updates the workload record the for workload matching workloadname according to
...
+GET - /api/racks/{rackid}
+Returns a single rack ID record.
+
+GET - /api/racks/rack-id/{blades}
+Get list of known blades in a rack.
+
+GET - /api/racks/rack-id/{TOR}
+Get Top of Racks details .
+
+GET - /api/racks/rack-id/{PDU}
```
Jim3Things, line 116: TOR and PDU are literals, not variables.  They shouldn't be in {..}

internal/services/frontend/inventory.go
```diff
@@ -0,0 +1,103 @@
...
+	// it there allows simple experimentation with just a browser.
+	//
+	// As a reminder,
+	//	 PUT is idempotent so translates to UPDATE in the CRUD methodolgy
+	//   POST is NOT idempotent so translates to CREATE in the CRUD methodolgy
+	//
+	routeInventory.HandleFunc("/{rackid}", handlerInventoryCreate).Methods("POST", "GET") // May be only GET
+	routeInventory.HandleFunc("/{rackid}", handlerInventoryRead).Methods("GET")
+	routeInventory.HandleFunc("/{rackid}", handlerInventoryUpdate).Methods("PUT", "GET")
+	routeInventory.HandleFunc("/{rackid}", handlerInventoryDelete).Methods("DELETE", "GET")
```
Jim3Things, line 60: We only need to support the GET method for now.

internal/services/frontend/doc.go
```diff
@@ -101,12 +101,37 @@ Updates the workload record the for workload matching workloadname according to
...
 operation code is invalid in some way.
 
+GET - /api/racks
+Get the list of all known racks.
+
+GET - /api/racks/{rackid}
+Returns a single rack ID record.
+
+GET - /api/racks/rack-id/{blades}
+Get list of known blades in a rack.
```
Jim3Things, line 111: We'll want

GET /api/racks/{rack-id}/blades 
to get the list of blades associated with that rack and

GET /api/racks/{rack-id}/blades/{blade-id}
to get details about a specific blade in a specific rack

internal/services/frontend/doc.go
```diff
@@ -101,12 +101,37 @@ Updates the workload record the for workload matching workloadname according to
...
+
 
 
 TODO
+
 /api/logs
+
 /api/stepper
 /api/injector
 /api/inventory
```
Jim3Things, line 133: The code has all the inventory under /api/inventory, but the examples above don't.  We just need to resolve whether /inventory or /blades is the root term

manpreet85:
> The code has all the inventory under /api/inventory, but the examples above don't.  We just need to resolve whether /inventory or /blades is the root term

manpreet85:
Can you please advise what is root term and how it impacts the code?

PR #17, opened 2020-04-06
Milestone 1 test, part 2

Jim3Things:
Add in initial logout impl, move trivial user definition into proto file, update tests


internal/services/frontend/users.go
```diff
@@ -86,18 +81,18 @@ func secret(w http.ResponseWriter, r *http.Request) {
     }
 
     // Print secret message
-    fmt.Fprintln(w, "secret message")
+    _, _ = fmt.Fprintln(w, "secret message")
```
MikeCJohnsonUk, line 84: Gotta ask, for curiosity, why the assignment of both output value to throw-away?

Jim3Things: It was just a lint fix.  The line is purely temporary and we don't care especially about whether it works (we obviously will with the real code).

PR #18, opened 2020-04-08
Add ability to run store tests using embedded etcd

MikeCJohnsonUk:
Add ability to tun store unit tests using an embedded instance of etcd. 
Re-enable store unit tests in run_all_uts.cmd
Also turns down log level for both the store unit tests and etcd to reduce the noise in the log
Fixes issue in ReadXxx() tests with error response for invalid key read attempt.

### DIFF
--- a/internal/clients/store/store_test.go
+++ b/internal/clients/store/store_test.go
@@ -44,18 +44,6 @@ func etcdStart() {
 	hostList[defaultEmbeddedEtcdNode] = struct{}{}
 	hostList[defaultEmbeddedEtcdAddr] = struct{}{}
 
-	/*
-			oldClient := embed.DefaultAdvertiseClientURLs
-			oldPeer := embed.DefaultInitialAdvertisePeerURLs
-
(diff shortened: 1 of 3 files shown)


internal/clients/store/store_test.go
```diff
@@ -4,28 +4,129 @@ package store
...
+	lcurl, _ := url.Parse(urlClient)
+	apurl, _ := url.Parse(urlPeer)
+	acurl, _ := url.Parse(urlClient)
+
+	etcdConfig.LPUrls = []url.URL{*lpurl}
+	etcdConfig.LCUrls = []url.URL{*lcurl}
+	etcdConfig.APUrls = []url.URL{*apurl}
+	etcdConfig.ACUrls = []url.URL{*acurl}
+
+	// Overrider the deagult log level "info" which is very noisy
```
Jim3Things, line 69: Overrider -> Override, deagult -> default

MikeCJohnsonUk: Sigh... fimble fungers....
I'll update and complete the PR without further review...

PR #19, opened 2020-04-09
User management updates, part 2

Jim3Things:
Completed operation support
Simple user storage class
Correct list operation
Completed read user detail
Updated UTs
Commented
Lint cleanup


pkg/protos/admin/users.proto
```diff
@@ -12,8 +12,40 @@ option go_package="github.com/Jim3Things/CloudChamber/pkg/protos/admin";
...
+  // The rights this user has
+  bool account_manager = 5;
+}
+
+// Extend the base user definition with the revision number.  This is
+// managed by the storage layer only, and is used as match criteria on
+// updates.
+message user_internal {
+  user user = 1;
+  int64 revision = 2;
```
MikeCJohnsonUk, line 31: So, does this mean this will not be seen in a message as such? If not, what are the rule for set/expected values?

internal/services/frontend/users.go
```diff
@@ -32,313 +35,585 @@ const (
...
+
+// Each user entry has an associated key which is the lowercased form of the
+// username.  The supplied name is retained as an attribute in order to present
+// the form that the caller originally used for display purposes.
+
+// DBUsers is a container used to established synchronized access to
 // the in-memory set of user records.
 //
-type DbUsers struct {
+type DBUsers struct {
```
MikeCJohnsonUk, line 53: Oh, this just grates. Every time I see it. Sorry.

internal/services/frontend/users.go
```diff
@@ -32,313 +35,585 @@ const (
...
-}
-
 // Process an http request for the list of users.  Response should contain a document of links to the
 // details URI for each known user.
 func handlerUsersList(w http.ResponseWriter, r *http.Request) {
-    _ = tr.WithSpan(context.Background(), methodName(), func(ctx context.Context) error {
-        span := trace.SpanFromContext(ctx)
+    _ = tr.WithSpan(context.Background(), methodName(), func(ctx context.Context) (err error) {
+        err = doSessionHeader(ctx, w, r, func(ctx context.Context, span trace.Span, session *sessions.Session) error {
+            return canManageAccounts(session, "") })
```
MikeCJohnsonUk, line 212: Definitely better. If _feels_ more comprehensible. It's at least easier to examine.

Still wish it was easier to override the opinionated indenting though...

pkg/protos/admin/users.pb.go
```diff
@@ -27,10 +27,13 @@ const _ = proto.ProtoPackageIsVersion3 // please upgrade the proto package
...
-	PasswordHash         []byte   `protobuf:"bytes,2,opt,name=password_hash,json=passwordHash,proto3" json:"password_hash,omitempty"`
-	UserId               int64    `protobuf:"varint,3,opt,name=user_id,json=userId,proto3" json:"user_id,omitempty"`
-	Enabled              bool     `protobuf:"varint,4,opt,name=enabled,proto3" json:"enabled,omitempty"`
+	// The username, held so as to preserve the original case
+	Name         string `protobuf:"bytes,1,opt,name=name,proto3" json:"name,omitempty"`
+	PasswordHash []byte `protobuf:"bytes,2,opt,name=password_hash,json=passwordHash,proto3" json:"password_hash,omitempty"`
+	UserId       int64  `protobuf:"varint,3,opt,name=user_id,json=userId,proto3" json:"user_id,omitempty"`
+	Enabled      bool   `protobuf:"varint,4,opt,name=enabled,proto3" json:"enabled,omitempty"`
+	// The rights this user has
+	AccountManager       bool     `protobuf:"varint,5,opt,name=account_manager,json=accountManager,proto3" json:"account_manager,omitempty"`
```
MikeCJohnsonUk, line 36: Not sure I understand what is meant by "AccountManager"?
Comment implies it's the set of rights/privs accorded to this identity?

Unless, is this meant to indicate this identity can manage accounts? In which case should this not be the "ManageAccounts" right?

internal/services/frontend/users.go
```diff
@@ -32,313 +35,585 @@ const (
...
+        PasswordHash: passwordHash,
+        Enabled: enabled,
+        AccountManager: accountManager})
+}
+
+func userRemove(name string) error {
+    return dbUsers.Remove(name)
+}
+
+func userVerifyPassword(name string, password []byte) error {
```
MikeCJohnsonUk, line 532: How careful do we need to be about having copies of the password string floating around in abandoned memory? Do we need to consider proactive clearing out, de-allocation etc? My understanding is that string are immutable. But since this is a []byte can we safely zero when finished with?

internal/config/settings.go
```diff
@@ -25,6 +25,7 @@ const (
     DefaultHost = ""
     DefaultRootFilePath = "."
     DefaultSystemAccount = "Admin"
+    DefaultSystemPassword = "SystemPassword"
```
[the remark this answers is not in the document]
Jim3Things: The yaml overrides, and I wanted the yaml values to be different so I could be sure that the override was working.

pkg/protos/admin/users.proto
Jim3Things: Not sure what the question is, so let me try to explain and you can let me know if this gets anywhere near your question.  user_internal is used only by the data access layer and the internal DB structure.  It includes the business logic version of the user structure (as field user) and a revision number.  That number is returned by Get and provided as a match tag on Update.  The actual new value is decided on by the data access layer.

So someone looking at the business logic version of user only would not see the revision number and there is no path that allows them to store a revision number value.

Does this help?

internal/services/frontend/frontend_test.go
```diff
@@ -72,62 +80,183 @@ func doHttp(req *http.Request) *http.Response {
...
+	response := doHttp(request)
+
+	body, err := ioutil.ReadAll(response.Body)
+	assert.Nilf(t, err, "Failed to read body returned from call to handler for route %v: %v", baseURI + admin, err)
+
+	t.Logf("[?op=login]: SC=%v, Content-Type='%v'\n", response.StatusCode, response.Header.Get("Content-Type"))
+	t.Log(string(body))
+
+	assert.Equal(t, 1, len(response.Cookies()), "Unexpected number of cookies found")
+	assert.Equal(t, http.StatusBadRequest, response.StatusCode, "Handler returned unexpected error: %v", err)
```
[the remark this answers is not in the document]
Jim3Things: good catch.  Fixed.

pkg/protos/admin/users.pb.go
Jim3Things: Yeah, can go with ManageAccounts.  Fixed

internal/services/frontend/users.go
Jim3Things: That would be good to do.  Let me add it as a TODO for now and research what the options are.

internal/services/frontend/users.go
Jim3Things: I'm going to assume that it doesn't grate enough to require changing, but yeah, I know the feeling.

internal/services/frontend/users.go
Jim3Things: I've tried a simple format change in the version I'm about to push.  Let me know if that helps.

internal/services/frontend/users.go
MikeCJohnsonUk: Indeed. It'll keep the lint-er happy at least.

Anyway, I've lived with way more offensive things before now.

pkg/protos/admin/users.pb.go
MikeCJohnsonUk: Thank you.

It took me a while after making the comment to figure out *why* it was bugging me - action vs. role.

internal/config/settings.go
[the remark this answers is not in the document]
MikeCJohnsonUk: Ok.

internal/services/frontend/frontend_test.go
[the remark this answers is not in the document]
MikeCJohnsonUk: Sorry to bug you, but it looks like there are multiple instances of the same problem further down.

pkg/protos/admin/users.proto
MikeCJohnsonUk: I think that answers. I was just wanting to assure myself that the revision field was not going to be (active?) in a message between two services.

internal/services/frontend/users.go
MikeCJohnsonUk: Ok

internal/services/frontend/frontend_test.go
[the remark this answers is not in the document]
Jim3Things: Fixed in update I'm about to push.

PR #21, opened 2020-04-17
Internal project structure update

Jim3Things:
Update and fill in several internal readme.md files that describe the subtrees in the project.
Some relocation and cleanup of .cmd files to rationalize the project structure.
Remove two directories that I'm sure will not be useful.


pkg/protos/README.md
```diff
@@ -1,5 +1,5 @@
 This subtree contains the various proto definitions used by Cloud Chamber.
 
-Both the proto and the generated .pb.go files are checked in.  If you make
-a change to a proto file you must first run build_protos.bat to update the
-generated files before building and committing.
\ No newline at end of file
+Both the proto and the generated go files are checked in.  If you make
```
MikeCJohnsonUk, line 3: I was meaning to ask, and this provides a place to do so, but why are we checking in generated files? Always leads to a bit of a surprise when "git status" reports a bunch of changes I hadn't made...

Jim3Things: Hmm.  Yeah.  When I started I was just using the GoLand build, which didn't know what to do with proto files.  That led me to checking in the generated files.  But now that we use buildall, we don't need to do that.  I think you're right and we can remove them - I'll remove them from this branch and add a .gitignore rule for them.

MikeCJohnsonUk: Works for me.

.gitignore
```diff
@@ -1,7 +1,6 @@
 .exe
 .vscode
-.pb.go
-.pg.validate.go
 /deployments/
+/pkg/protos/**/*.go
```
MikeCJohnsonUk, line 4: Is the double asterisk significant?

Jim3Things: Yes it is.  /**/ means zero or more directories - it's what causes this rule to apply to all the subdirectories under pkg/protos

MikeCJohnsonUk: Ok.

PR #22, opened 2020-04-19
User update, delete, unit tests

Jim3Things:
This change includes support for update & delete operations on users, restructures and extends the unit tests

### DIFF
--- a/internal/services/frontend/frontend.go
+++ b/internal/services/frontend/frontend.go
@@ -161,6 +161,13 @@ func (he *HTTPError) Error() string {
     return he.Base.Error()
 }
 
+func NewErrBadMatchType(match string) *HTTPError {
+    return &HTTPError{
+        SC : http.StatusBadRequest,
+        Base: fmt.Errorf("CloudChamber: match value %q is not recognized as a valid integer", match),
+    }
+}
+
 func initHandlers() error {
 
     server.handler = mux.NewRouter()
--- a/internal/services/frontend/frontend.go
+++ b/internal/services/frontend/frontend.go
@@ -91,11 +91,6 @@ var (
     //
     ErrUserAuthFailed = errors.New("CloudChamber: authentication failed, invalid user name or password")
 
-    // ErrUserPermissionDenied indicates the user does not have the appropriate
-    // permissions for the requested operation.
-    //
-    ErrUserPermissionDenied = errors.New("CloudChamber: permission denied")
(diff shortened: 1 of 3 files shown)


internal/services/frontend/frontend.go
```diff
@@ -161,6 +126,67 @@ func (he *HTTPError) Error() string {
     return he.Base.Error()
 }
 
+// +++ HTTPError specializations
+
+// ErrNoLoginActive indicates that the specified user is not logged into this session
+func NewErrNoLoginActive(name string) *HTTPError {
```
MikeCJohnsonUk, line 132: The significance of the "New" prefix on this and the following errors?

internal/services/frontend/users.go
```diff
@@ -319,34 +329,111 @@ func handlerUsersRead(w http.ResponseWriter, r *http.Request) {
...
+            return err
+        }
+
+        // Finally, get the current user definition and return it.  Again, protecting
+        // against the user getting deleted in another (racing) thread.
+        u, rev, err := dbUsers.Get(username)
+        if err != nil {
+            httpError(ctx, span, w, err)
+            return err
+        }
```
MikeCJohnsonUk, line 391: If there is always (likely) to be a Get() after an update, perhaps the update should return the newly updated user record on success? That way we 
1. only need a single call,
2. get rid of a potential error case (update succeeds, get fails)

internal/services/frontend/frontend.go
Jim3Things: I was following a pattern that I'd seen - the NewXxx creates a new object for Xxx.  In this case, the ErrNoLoginActive is a specialization of HTTPError, which makes it a bit funky.

internal/services/frontend/users.go
Jim3Things: Yeah, I thought about that, and probably should go that way.

Jim3Things: Fixed in this latest update.

internal/services/frontend/frontend.go
MikeCJohnsonUk: Hhmm, I find myself conflicted. I see the reasoning, but it still looks wrong somehow. I need to think on this to figoure out how I need to reconcile those two. But no reason to hold the checkin.

internal/services/frontend/users.go

PR #26, opened 2020-04-25
Initial cut at bash equivalents for the Windows utility scripts

MikeCJohnsonUk:
These are a first cut of bash shell scripts to provide similar functionality to their similarly names Windows CMD file equivalents.

Except testall.sh which is the moral equivalent of run_all_uts.cmd


build/dev_tools/fetchall.sh
```diff
@@ -0,0 +1,29 @@
...
+go get github.com/golang/protobuf
+go get github.com/golang/protobuf/protoc-gen-go
+go get github.com/envoyproxy/protoc-gen-validate
+
+go get github.com/gorilla/mux
+go get github.com/gorilla/securecookie
+go get github.com/gorilla/sessions
+
+#go get github.com/davecgh/go-spew/spew
+#go get github.com/pmezard/go-difflib/difflib
```
Jim3Things, line 14: Why are these commented out?  If we don't need them, then should they be removed from both forms?

build/dev_tools/updateall.sh
```diff
@@ -0,0 +1,27 @@
...
+go get -u github.com/golang/protobuf
+go get -u github.com/golang/protobuf/protoc-gen-go
+go get -u github.com/envoyproxy/protoc-gen-validate
+
+go get -u github.com/gorilla/mux
+go get -u github.com/gorilla/securecookie
+go get -u github.com/gorilla/sessions
+
+#go get -u github.com/davecgh/go-spew/spew
+#go get -u github.com/pmezard/go-difflib/difflib
```
Jim3Things, line 14: Same comment here.

MikeCJohnsonUk: Same response.

build/dev_tools/fetchall.sh
MikeCJohnsonUk: I thought these were left over from things I was experimenting with/planned to possibly use way back. However, even if commented out, they are fetched/updated as a dependency.

I'll re-instate them by un-commenting..

PR #27, opened 2020-04-26
Fix inadvertent file deletion bug

MikeCJohnsonUk:
If GOPATH is not defined, the cd /d fails but is then followed by a unilateral del *.* which can lead to the deletion of all files in the *current* directory, rather than the deployment directory.

Fix is to check that the deployments directory based on a GOPATH env var actually exists before trying to use it. It might be the wrong deployments directory if there are multiple copies of the repo, but at least it's not deleting a bunch of arbitrary files.

### DIFF
--- a/build/cleanall.cmd
+++ b/build/cleanall.cmd
@@ -1,7 +1,22 @@
-pushd .
+@rem
+@rem
+@rem    C L E A N A L L . C M D
+@rem
+@rem
+@rem
 
-cd /d %GOPATH%\src\github.com\Jim3Things\CloudChamber\deployments
-del *.*
+@if /i "%DbgScript%" == "" @echo off
 
-popd
+if exist "%GOPATH%\src\github.com\Jim3Things\CloudChamber\deployments" (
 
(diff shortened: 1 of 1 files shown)


build/cleanall.cmd
```diff
@@ -1,7 +1,22 @@
...
-cd /d %GOPATH%\src\github.com\Jim3Things\CloudChamber\deployments
-del *.*
+@if /i "%DbgScript%" == "" @echo off
 
-popd
+if exist "%GOPATH%\src\github.com\Jim3Things\CloudChamber\deployments" (
 
+  pushd "%GOPATH%\src\github.com\Jim3Things\CloudChamber\deployments"
+
+  del *.*
```
Jim3Things, line 14: Can you make this a del /q, or whatever to avoid the confirmation question?  It was a minor thing that I was going to put on my list and since you're in there...

MikeCJohnsonUk: Certainly. I wasn't sure whether or not the prompt was intentional.

I'll just check-in with that change without re-review. Also, I'll make a similar change to the UNIX shell script when I'm next in there.

PR #28, opened 2020-04-28
Add Windows build environment doc

MikeCJohnsonUk:
Added the Windows build environment setup instructions as a markdown file.

This is and updated version of the instructions I sent in mail a few days back. These original version was moved to a Microsoft Word .docx file and then converted to markdown using the pandoc utility. The VsCode previewer seems to be quite happy with the result.

### DIFF
--- a/docs/BuildEnvironmentSetupForWindows.md
+++ b/docs/BuildEnvironmentSetupForWindows.md
@@ -18,6 +18,36 @@ The environment will make use of the following
 VsCode is not required and other IDEs might be preferred. Configuring
 those IDEs is left as an exercise for the reader.
 
+# Assumptions
+
+There are some paths required with associated environment variables.
+Throughout the remainder of this documents these will be assumed to have
+the following values
+
+  - These instructions assume that the Go environment is rooted at the
+    C:\\Chamber directory and that the value of the environment variable
+    GOPATH is set to C:\\Chamber. This directory is an arbitrary choice
+    and an alternative location on a local filesystem can be chosen
+    instead of C:\\Chamber to suit the system being used provided GOPATH
+    is set to the same location.
+
+Note, avoid using a network share for the local copy of the repository
+as some problems have been encountered for files which originated on a
+non-Windows system.
+
+  - These instructions install some tools and assume these are placed in
+    the C:\\GoTools directory. This is an arbitrary selection and any
+    alternative location can be selected to suite the system being used.
+    If an alternative location is used, substitute the selected location
+    for any reference to C:\\GoTools.
+
+  - These instructions assume the Windows system is installed on the C:
+    drive and standard installation locations are used. In particular,
+    the instructions assume that the value of the standard environment
+    variable ProgramFiles has the value “C:\\Program Files”. If an
+    alternative value is used, substitute this value wherever the
+    instructions refer to “C:\\Program Files”.
+
 # Windows Basics
 
 ## Windows Command prompt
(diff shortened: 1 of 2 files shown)


docs/BuildEnvironmentSetupForWindows.md
```diff
@@ -0,0 +1,453 @@
...
+> git
+
+The utility should supply the top-level help. If not, git is probably
+not correctly installed. Either a repair can be attempted, but it is
+generally simpler to uninstall it and then re-install it (recommended).
+
+To see if it can be simply repaired, at the Command prompt check the
+value of the PATH environment variable verify that it contains
+
+> C:\\Program Files\\Git\\cmd
```
Jim3Things, line 134: I'm not sure the best way to handle this, but how would you want to handle cases where the installation path isn't C: based?  Maybe a note somewhere about how C:\ is the default, but your system may be different?

docs/BuildEnvironmentSetupForWindows.md
```diff
@@ -0,0 +1,453 @@
...
+> git clone <https://github.com/Jim3Things/CloudChamber>
+> C:\\Chamber\\src\\github.com\\Jim3Things\\CloudChamber
+
+or the command can be broken down into multiple steps using
+
+> mkdir C:\\Chamber\\src\\github.com\\Jim3Things\\CloudChamber
+> 
+> pushd C:\\Chamber\\src\\github.com\\Jim3Things
+> 
+> git clone <https://github.com/Jim3Things/CloudChamber>
```
Jim3Things, line 254: I'm not sure about these directions - I think you'd want to use %GOPATH% as part of the directory.  What I've seen is that if the repo is outside the %GOPATH% destination (%GOPATH%\src\github.com\Jim3Things\CloudChamber).  The instructions above sort of assume that C:\Chamber is GOPATH.

docs/BuildEnvironmentSetupForWindows.md
```diff
@@ -0,0 +1,453 @@
...
+This will initialize a git repository and populate it with the
+CloudChamber files under the specified directory.
+
+If using the go tool, set the GOPATH environment variable to the root of
+the Go project directory and then use the go tool to fetch the initial
+copy of the repository, e.g.
+
+> set GOPATH=C:\\Chamber
+> 
+> go get github.com/Jim3Things/CloudChamber
```
Jim3Things, line 265: Does this actually do the same thing?  Can I issue a branch creation and push that branch or do a PR from a go get command?  Just wondering as I don't know.

docs/BuildEnvironmentSetupForWindows.md
MikeCJohnsonUk: Well, that is increasingly unlikely these days, but yeah, this probably should be based on the standard env vars. I'll update the doc.

docs/BuildEnvironmentSetupForWindows.md
MikeCJohnsonUk: I'd missed the z: ref. Rats.

But there arguably is a little bit of a chicken and egg situation here. However, it should be possible to set an assumption and then be consistent throughout.

I'll do a consistency pass.

MikeCJohnsonUk: Also, I think adding an assumption section to define some of these base terms and expectation might work for defining a couple of these "defaults"

docs/BuildEnvironmentSetupForWindows.md
MikeCJohnsonUk: Ah, what? My understanding is that go get is effectively the same as git clone with suitable changes for the path package name. Not sure where branch creation or PR come from here.

You can certainly do a branch creation from git but I've not seen any reference to an equivalent for "go <xxx>".

As far as I can see, a PR is purely a github thing layered at the server on top of some git magic to pushed branches and a merge.

MikeCJohnsonUk: Further on the go get vs. git clone

I don't know if these are always the same thing. It likely depends on whatever is stored at the source. In our case, yes they seem to be directly equivalent. If go get is used, subsequent git command are completely happy.

Jim3Things: Ok, I got a chance to verify that the the go get produces what looks like a cloned repo.  I'm going to assume that this works for now.

docs/BuildEnvironmentSetupForWindows.md
Jim3Things: The changes above really do help.

docs/BuildEnvironmentSetupForWindows.md
Jim3Things: Yes, that change really did help a lot.

PR #29, opened 2020-05-03
Add version flag to executables

MikeCJohnsonUk:
Adds a -version flag to each of the executables in which the build script will place useful build-time defined/obtained information.

This works by using the ldflags -X option to replace the value of a variable at the moral equivalent of link time. The name of the symbol must be known and the names chosen ensure that this requirement holds true.

This looks a little hokey, but I've not so far found anything more robust. Oh, for compile time constants and/or resource files...

### DIFF
--- a/build/buildall.cmd
+++ b/build/buildall.cmd
@@ -28,7 +28,7 @@
 @rem
 @if " " == "%UpdateHour:~0,1%" set UpdateHour=0%UpdateHour:~1,1%
 
-@set UpdateDateTime=%UpdateYear%%UpdateMonth%%UpdateDay%-%UpdateHour%%UpdateMinute%%UpdateSecond%
+@set BuildDateTime=%UpdateYear%%UpdateMonth%%UpdateDay%-%UpdateHour%%UpdateMinute%%UpdateSecond%
 
 
 @rem Update PATH to include the tools from the local Go environment. This is a temporary update
@@ -41,7 +41,15 @@
 @set CCDeployments=%CCRoot%\deployments
 
 
-pushd %gopath%\src
+@set BUILD_VERSION=v0.0.1
+
+@for /F "usebackq delims=" %%i in (`git symbolic-ref --short HEAD`)        do @set BUILD_BRANCH=%%i
+@for /F "usebackq delims=" %%i in (`git log -n 1 --pretty^=format:"%%cI"`) do @set BUILD_BRANCH_DATE=%%i
+@for /F "usebackq delims=" %%i in (`git log -n 1 --pretty^=format:"%%H"`)  do @set BUILD_BRANCH_HASH=%%i
+
+@set LD_FLAGS="-X 'main.version=%BUILD_VERSION%' -X 'main.buildDate=%BuildDateTime%' -X 'main.buildBranch=%BUILD_BRANCH%' -X 'main.buildBranchDate=%BUILD_BRANCH_DATE%' -X 'main.buildBranchHash=%BUILD_BRANCH_HASH%'"
+
+@pushd %gopath%\src
 
 protoc --go_out=. --validate_out=lang=go:. github.com\Jim3Things\CloudChamber\pkg\protos\admin\users.proto
 protoc --go_out=. --validate_out=lang=go:. github.com\Jim3Things\CloudChamber\pkg\protos\common\capacity.proto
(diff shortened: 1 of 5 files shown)


MikeCJohnsonUk:
I've just realised there's a spaces/tabs substitution issue in just one of the source files. I'll try and sort that out without changing anything else.

build/buildall.cmd
```diff
@@ -41,7 +41,15 @@
...
 
-pushd %gopath%\src
+@set BUILD_VERSION=v0.0.1
+
+@for /F "usebackq delims=" %%i in (`git symbolic-ref --short HEAD`)        do @set BUILD_BRANCH=%%i
+@for /F "usebackq delims=" %%i in (`git log -n 1 --pretty^=format:"%%cI"`) do @set BUILD_BRANCH_DATE=%%i
+@for /F "usebackq delims=" %%i in (`git log -n 1 --pretty^=format:"%%H"`)  do @set BUILD_BRANCH_HASH=%%i
+
+@set LD_FLAGS="-X 'main.version=%BUILD_VERSION%' -X 'main.buildDate=%BuildDateTime%' -X 'main.buildBranch=%BUILD_BRANCH%' -X 'main.buildBranchDate=%BUILD_BRANCH_DATE%' -X 'main.buildBranchHash=%BUILD_BRANCH_HASH%'"
+
```
Jim3Things, line 51: Since this only works if you build using the script, can I suggest an alternative approach?  How about having this script write a .go file into a known (and .gitignore'd location) in the build tree.  Have that module export a 'GetBuildVersion() string' function that each main.go then calls.  That seems like it might be cleaner, and it may be easier to add to.

MikeCJohnsonUk: The problem with that approach though is that the version info is then linked to the time that .go file was written, not to the time the current binary is being built. If you remember to always build the one and then build your bin that too works, but if not...

I'm currently taking a look at go generate, but I think it suffers from the same problem. I've not seen anything so far which allows for a module-build-time value to be captured which is independent of either an external invocation (i.e. buildall.cmd) or a separately invoked utility.

So far at least, there doesn't seem to be a good answer.

Jim3Things:
Yup, all these approaches require that you use the buildall shell or
something like it... Even go generate isn't standalone.  Sigh.

Jim.

build/buildall.cmd
Jim3Things: I also can't find a good answer.  the '-X' flags approach does seem to be common.  I'll agree here, but there is one thing that bugs me a bit.  Right now we have repeated code in every main.go.  It'd be great if the actual code to dump the build data could be in a common function that gets invoked.  Not sure how much it would help, but a suggestion anyway.

cmd/controllerd/main.go
```diff
@@ -16,13 +16,31 @@ import (
...
+var (
+	// declare a bunch of variables whose values given here will be replaced by some
+	// build-time defined values. If for some reason those build-time replacements
+	// do not take place then the -version flag will dump these values as-is.
+	//
+	version         = "development"
+	buildDate       = "buildDate"
+	buildBranch     = "branch"
+	buildBranchHash = "branch-hash"
+	buildBranchDate = "branch-date"
```
Jim3Things, line 28: 'development' sort of hints at this, but it would be great if the 'show version' output could say explicitly that this is from a non-project ('non-official') build.

cmd/controllerd/main.go
```diff
@@ -16,13 +16,31 @@ import (
...
+	showVersion := flag.Bool("version", false, "display the current version of the program")
 	flag.Parse()
 
+	if *showVersion {
+		fmt.Printf("Version: %v\nBuildDate: %v\nBranch: %v (%v \\ %v)\n", version, buildDate, buildBranch, buildBranchDate, buildBranchHash)
+		os.Exit(0)
+	}
+
+	setup.Init(exporters.StdOut)
+
```
Jim3Things, line 43: As a+1, having the version data output at the start of the log stream would be excellent here.

build/buildall.cmd
MikeCJohnsonUk: I'll check this in as is for the present, but expect a follow-up where the version info is moved to a single module which is then linked in as required to the individual exe main() routines. That should at least keep all the icky-ness to a single location.

cmd/controllerd/main.go
MikeCJohnsonUk: Yup, I like that notion. I'll do that in the follow-on.

cmd/controllerd/main.go
MikeCJohnsonUk: Makes sense. I'll do this in the follow-on.

PR #30, opened 2020-05-04
Tracing improvements

Jim3Things:
I split off the tracing improvement from the actor work and cleaned it up a bit.

Add stack trace, labels, severity
Cleaned up case where I was redundantly setting the span status to OK

### DIFF
--- a/internal/tracing/server/tracing.go
+++ b/internal/tracing/server/tracing.go
@@ -14,10 +14,15 @@ import (
 	"google.golang.org/grpc/metadata"
 
 	"github.com/Jim3Things/CloudChamber/internal/tracing"
+	"github.com/Jim3Things/CloudChamber/pkg/protos/log"
 )
 
 // Interceptor intercepts and extracts incoming trace data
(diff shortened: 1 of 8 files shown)


internal/tracing/server/tracing.go
```diff
@@ -32,47 +37,83 @@ func Interceptor(ctx context.Context, req interface{}, info *grpc.UnaryServerInf
...
 
 	return handler(ctx, req)
 }
 
-func AddEvent(ctx context.Context, msg string, tick int64, reason string) {
+// +++ Exported trace invocation methods
+
+// There should be an Xxx and Xxxf method for every severity level, plus some
+// specific scenario functions (such as OnEnter to log an information entry
+// about arrival at a specific method).
```
MikeCJohnsonUk, line 51: There being an OnEnter() does lead to the question as to whether or not there would be a matching OnExit()? One assume yes?

Jim3Things: If we end up needing it, yes.

PR #31, opened 2020-05-08
Change version info generation to avoid using link-time symbol substitution 

MikeCJohnsonUk:
Removed the icky link-time symbol editing to inject versioning information into the binaries and replace it by adding a program (generate.go) which is run *at build-time* and gathers all the version information required and then uses that to generate a go module which is then included in a new package "version". This version package include a few access functions which are called from the binaries to print / retrieve / trace the version information.

This mechanism is readily extensible to gather whatever information is desired to be added should we choose to add more later on.

There are some changes we might wish to contemplate, for example, the "go generate" mechanism is completely superfluous, or all of the version package could be moved into the generated file but at least for now this works.

I'm interested in hearing what folks think.


pkg/version/version.go
```diff
@@ -0,0 +1,56 @@
...
+	"fmt"
+
+	"go.opentelemetry.io/otel/api/global"
+	"go.opentelemetry.io/otel/api/trace"
+)
+
+//go:generate go run generator\generate.go
+
+// ToString is a function which returns the version information formatted as a
+// single line, suitable for dumpting to a log file.
```
Jim3Things, line 17: typo: dumpting -> dumping

pkg/version/version.go
```diff
@@ -0,0 +1,56 @@
...
+//
+func ToString() string {
```
Jim3Things, line 19: Can this be local (e.g. toString)?

If you can make a type of the original struct, then I believe that a func (s *that_struct) String() string might be a better signature - that would let you use '%v' to format the full struct in the log and show cases.  As a thought.

pkg/version/version.go
MikeCJohnsonUk: Dumpting? I wonder what that is. Sounds interesting...

Done.

pkg/version/version.go
MikeCJohnsonUk: Yes, it could be a local. Originally planned on having the clients of the package access / trace the string.

Interesting notion about the direct string-ification of the struct. Should have thought of that. I'll experiment. It could be useful for the trace, but I'm not so sure about the Show() output as that is more "formatted".

MikeCJohnsonUk: Ok, changed my mind.

Mainly because the trace and show variants use different formatting, so nothing really to be collapsed. And although the version info could be a struct, it's only used internally to the package and so making it a struct adds a bunch of characters but doesn't much help.

If things change, and the info needs to be public, then it should be made into a struct, but for now I'm not going to bother.

I will make ToString() local though, and I'm changing the name of TraceVersion() to just trace:  it's been bugging me!

MikeCJohnsonUk: That should be TraceVersion() ==> Trace()

PR #41, opened 2020-05-19
Simplify the context tracking in the actors

Jim3Things:
Move all context state extensions to attributes on context.Context, and track appropriately.
Remove duplication MethodName implementation.

### DIFF
--- a/internal/services/stepper_actor/sm.go
+++ b/internal/services/stepper_actor/sm.go
@@ -10,7 +10,6 @@ import (
     "github.com/AsynkronIT/protoactor-go/scheduler"
     "github.com/golang/protobuf/ptypes"
     "github.com/golang/protobuf/ptypes/empty"
-    trc "go.opentelemetry.io/otel/api/trace"
 
     "github.com/Jim3Things/CloudChamber/internal/sm"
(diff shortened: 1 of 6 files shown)


internal/services/stepper_actor/sm.go
```diff
@@ -293,25 +297,25 @@ func (act *Actor) HandleDelay(c context.Context, span trc.Span, ctx actor.Contex
...
         slot := value.([]*actor.PID)
-        slot = append(slot, ctx.Sender())
+        slot = append(slot, ca.Sender())
         act.waiters.Put(due, slot)
     }
 
     // Now, check for any waiters that have already expired (this handles the
     // case where the delay due time had already passed)
-    act.checkForExpiry(c, span, ctx)
+    act.checkForExpiry(ctx)
```
MikeCJohnsonUk, line 311: Just checking, this really is still ctx rather than ca? I'm just going on a pattern match here...

Jim3Things: Yes, it is.  I fixed the names while doing the rest, which does make the deltas look a bit confusing.

PR #42, opened 2020-05-20
Move users support to common tracing code

Jim3Things:


internal/services/frontend/users.go
```diff
@@ -359,41 +355,42 @@ func handlerUsersUpdate(w http.ResponseWriter, r *http.Request) {
...
         }
 
         // Now with the input in hand, get the user (if it exists).
-        _, rev, err := dbUsers.Get(username)
+        _, _, err = dbUsers.Get(username)
         if err != nil {
-            httpError(ctx, span, w, err)
+            httpError(ctx, w, err)
             return err
         }
```
MikeCJohnsonUk, line 374: Isn't this now pointless? Given that the update has to handle a non-existent user anyway, shouldn't this just be folded into the userUpdate() call, especially since the revision isn't being passed in anyway? 

I'm happy for it to stay, it's just that I'm not sure I see the value.

Jim3Things: Argh.  Thanks for pointing that out.  Shows that I was rushing.  Yes, it should just go away.  Fixed now.

PR #49, opened 2020-06-06
Refactor common test support out of the user test

Jim3Things:


internal/services/frontend/common_test.go
```diff
@@ -0,0 +1,148 @@
...
+
+// Common test startup method.  This is the _only_ Test* function in this
+// file.
+func TestMain(m *testing.M) {
+    commonSetup()
+
+    os.Exit(m.Run())
+}
+
+// Estalish the test environment, including starting a test frontend service
```
MikeCJohnsonUk, line 52: Establish generally includes a 'b'

internal/services/frontend/users_test.go
```diff
@@ -54,84 +43,6 @@ var (
...
-}
-
-// Get the body of a response, unmarshaled into the supplied message structure
-func getJsonBody(resp *http.Response, v proto.Message) error {
-	defer func() { _ = resp.Body.Close() }()
-	return jsonpb.Unmarshal(resp.Body, v)
-}
-
 // Log the specified user into CloudChamber
 func doLogin(t *testing.T, user string, password string, cookies []*http.Cookie) *http.Response {
```
MikeCJohnsonUk, line 47: Just a question, but should perhaps doLogin() and the associate doLogout() also move to the common_test module? Do we envisage that it might be more universally useful/needed later on?

internal/services/frontend/users_test.go
```diff
@@ -266,27 +170,27 @@ func TestLoginSessionRepeat(t *testing.T) {
...
-	response := doLogin(t, "Admin", adminPassword, nil)
+	response := doLogin(t, randomCase(adminAccountName), adminPassword, nil)
 
 	assert.Equal(t, 1, len(response.Cookies()), "Unexpected number of cookies found")
 	assert.Equal(t, http.StatusOK, response.StatusCode, "Handler returned unexpected error: %v", response.StatusCode)
 
 	// ... and logout, which should succeed
 	//
-	response = doLogout(t, "admin", response.Cookies())
+	response = doLogout(t, randomCase(adminAccountName), response.Cookies())
```
MikeCJohnsonUk, line 180: Do we want to have explicit testing for using names distinguished only by case as separate from identical names, even if just for a few basic early on tests.. It's just that I had a little flag being raised on seeing a likely different name used for the logout. 

Actually, should we even be using the name for the logout rather than (say) a session token?

Jim3Things: With this change I didn't.  I can add a test around 'Alice' that specifically tests for different case.  I can also move the randomCase call inside doLogin & doLogout.

Fwiw, the actual logout code does use the session.  The API takes the username for symmetry with ?op=login, so that the client can use the same basic address for both.  Is that goodness?  <shrug>  Don't know.  But since the path contains a username there is a check to verify that the supplied username is (case insensitive) the same as the username used to log in with.  Fwiw.

My current thought is to leave ?op=logout where it is, to make sure there's a specific alice case test, and to leave the rest of the changes as-is.

internal/services/frontend/users_test.go
Jim3Things: Yeah, makes sense.  Don't need it yet, but might as well move it out now.

MikeCJohnsonUk: Ok.

internal/services/frontend/users_test.go
MikeCJohnsonUk: Ok

PR #54, opened 2020-06-08
RackList test08Jun20

manpreet85:
MSM08Jun20: Created an array that captures the body response. 

MSM08Jun20: A test was added to check if all the racks were uniquely listed in the array.

### DIFF
--- a/internal/services/frontend/DBInventory.go
+++ b/internal/services/frontend/DBInventory.go
@@ -66,7 +66,7 @@ func InitDBInventory() error {
 	return nil
 }
 
-// Scan the set of known users in the store, invoking the supplied
+// Scan the set of known blades the store, invoking the supplied
 // function with each entry.
 func (m *DBInventory) Scan(action func(entry string) error) error {
 	m.Mutex.Lock()
--- a/internal/services/frontend/DBInventory.go
+++ b/internal/services/frontend/DBInventory.go
@@ -10,29 +10,17 @@ package frontend
 
 import (
 	"sync"
+
+	common "github.com/Jim3Things/CloudChamber/pkg/protos/common"
+	pb "github.com/Jim3Things/CloudChamber/pkg/protos/inventory"
 )
-
-type Tor struct {
-}
-
-type Pdu struct {
-}
-
-type blade struct {
-}
-
-type rack struct {
-	tor    Tor
-	pdu    Pdu
-	blades map[int64]*blade
-}
 
 // DBInventory is a container used to establish synchronized access to
 // the in-memory set of Racks records.
-//
+
 type DBInventory struct { //A struct for the collection of Racks
 	Mutex sync.Mutex
-	Racks map[string]*rack
+	Racks map[string]*pb.ExternalRack
 }
 
 var dbInventory *DBInventory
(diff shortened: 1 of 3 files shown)


internal/services/frontend/DBInventory.go
```diff
@@ -65,3 +65,18 @@ func InitDBInventory() error {
 
 	return nil
 }
+
+// Scan the set of known users in the store, invoking the supplied
+// function with each entry.
```
Jim3Things, line 70: Known blades

internal/services/frontend/DBInventory_test.go
```diff
@@ -21,7 +24,20 @@ func TestListRacks(t *testing.T) {
 	body, err := getBody(response)
 
 	log.Printf("ListRacks: SC=%v, Content-Type='%v'\n", response.StatusCode, response.Header.Get("Content-Type"))
-	log.Println(string(body))
+	// log.Println(string(body))
+	s := string(body)
+	var splits = strings.Split(s, "\n")
+	fmt.Println(splits)
+	assert.Equal(t, splits, "[Racks (List) /api/rasks/rack1 /api/racks/rack2 ]", "Error found")
```
Jim3Things, line 31: This isn't quite what we need.  This form basically replaces the '\n' char with ' ' by splitting it and then using the default array formatting to list each entry with a space.  That will work, iff rack1 is before rack2 in the output list.

I think we want something that follows the pseudo-code below more directly.  One hint is that the entry checks are probably in two parts, i.e. something like (for rack 1):
found1 := splits[1] == "/api/racks/rack1" || splits[2] == "/api/racks/rack1"
assert.True(t, found1, "rack1 was not found)

Also, there's a typo in the test literal above - rack1 has 'rasks' rather than 'racks'.  Has the test been run?

manpreet85: The test ran resulting in error. but the print ln function listed the racks.

I will make these changes tonight

internal/services/frontend/DBInventory_test.go
```diff
@@ -28,9 +28,13 @@ func TestListRacks(t *testing.T) {
 	s := string(body)
 	var splits = strings.Split(s, "\n")
 	fmt.Println(splits)
+	title := splits[0] == "Racks (List) "
 	found1 := splits[1] == "/api/racks/rack1" || splits[2] == "/api/racks/rack1"
+	found2 := splits[1] == "/api/racks/rack2" || splits[2] == "/api/racks/rack2"
 
+	assert.Equal(t, title, "Racks (List)")
 	assert.Equal(t, found1, "rack 1 was not found")
+	assert.Equal(t, found2, "rack2 was not found")
```
Jim3Things: Sorry I missed this before, but assert.Equal should be assert.True for the found1 & found2 tests.  Also, just for consistency, the associated messages should either start with "rack1" & "rack2" or "rack 1" & "rack 2".

manpreet85: I thought that we are testing if in the array the listed racks are rack1 @ array[1] and rack2 @ array[2]. or vice versa. 
Now I am confused. can you please explain the your comments this bit "Also, just for consistency, the associated messages should either start with "rack1" & "rack2" or "rack 1" & "rack 2""

manpreet85: ?

Jim3Things: Sorry - I was a bit terse there.  The substantive part of the comment is that you're using assert.Equals rather than assert.True.  I'm not quite sure why it is working, as it should be comparing the result (found1 or found2) to the string.  Since the found1 and found2 are booleans, you should use assert.True.

The second part was a nit, and worth mentioning only because there was already a reason to change the lines.  It really only has to do with the formatting of the text string - in line 36 the test prints out 'rack 1 was not found' while line 37 has 'rack2 was not found'.  It would be good if they used consistent spacing - i.e. either change line 36 to 'rack1 was not found' or line 37 to 'rack 2 was not found'.

hope this clears up my comment.

PR #56, opened 2020-06-11
Get Store to use standard config files

MikeCJohnsonUk:
This is step one in a sequence of store related changes. 

This step is to make the store pay attention to the standard config file. At present it still refers to the embedded etcd instance but that will go away completely in step 2. (I wanted to get this checked in to unblock the tests)

1. The store (according to the config file) will now use an external etcd instance, so scripts\startetcd.cmd should be run prior to starting any store related tests.
2. The store tests place use a sub-tree namespace to the main CloudChamber store. This will prevent any interference between the main store and any test data. Given a single store, we may need to go further and have a specific namespace per test-instance if we want to allow multiple test to run simultaneously.
3. The store tests now run cleanly on the current etcd version.

Upcoming in future PRs
step 2 - remove embedded etcd altogether
step 3 - convert to standard tracing (currently blocked by embedding of etcd)

### DIFF
--- a/internal/clients/store/store.go
+++ b/internal/clients/store/store.go
@@ -71,8 +71,11 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"strings"
 	"sync"
 	"time"
+
+	"github.com/Jim3Things/CloudChamber/internal/config"
 
 	"go.etcd.io/etcd/clientv3"
 	"go.etcd.io/etcd/clientv3/namespace"
@@ -81,15 +84,15 @@ import (
 	"google.golang.org/grpc/status"
 )
 
+// All CloudChamber key-value pairs are stored under a common namespace root.
+// This is fixed and cannot be changed programatically. However, to help
+// certain situations, such as test, an additional suffix to the namespace
+// can be added, changed, removed etc.
+//
+// At no time can the root namespace be removed.
+//
 const (
-	defaultEndpointNode = string("localhost")
-	defaultEndpointPort = string("2379")
-	namespacePrefix     = string("/CloudChamber/v0.1")
-	defaultEndpoint     = defaultEndpointNode + ":" + defaultEndpointPort
-
-	defaultTimeoutConnect = 5 * time.Second
-	defaultTimeoutRequest = 5 * time.Second
-	defaultTraceFlags     = traceFlagEnabled /*| traceFlagExpandResults */
+	cloudChamberNamespace = string("/CloudChamber/v0.1")
 )
 
 // TraceFlags is the type used when setting of fetching the relevant flags
@@ -105,13 +108,14 @@ type global struct {
 	Stores     map[string]Store
 	StoreMutex sync.Mutex
 
-	DefaultTimeoutConnect time.Duration
-	DefaultTimeoutRequest time.Duration
-	DefaultEndpoints      []string
-	DefaultTraceFlags     TraceFlags
+	DefaultTimeoutConnect  time.Duration
+	DefaultTimeoutRequest  time.Duration
+	DefaultEndpoints       []string
+	DefaultTraceFlags      TraceFlags
+	DefaultNamespaceSuffix string
 }
 
-// Store is a struct used to collect al the interesting state value associated with
+// Store is a struct used to collect all the interesting state values associated with
 // communicating with an instance of the store
 //
 type Store struct {
@@ -123,6 +127,7 @@ type Store struct {
 	UnprefixedKV      clientv3.KV
 	UnprefixedWatcher clientv3.Watcher
 	UnprefixedLease   clientv3.Lease
+	NamespaceSuffix   string
 	TraceFlags        TraceFlags
 }
 
@@ -145,8 +150,6 @@ type KeyValueResponse struct {
 
 var (
 	storeRoot global
-
-	defaultEndpoints = []string{defaultEndpoint}
 
 	// ErrStoreUnableToCreateClient indicates that it is not currently possible
 	// to create a client.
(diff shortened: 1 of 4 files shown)


internal/clients/store/store.go
```diff
@@ -411,6 +430,37 @@ func (store *Store) SetTimeoutRequest(timeout time.Duration) (err error) {
...
 func (store *Store) GetTimeoutRequest() time.Duration { return store.TimeoutRequest }
 
+// SetNamespaceSuffix is a method that can be used to update the namespace prefix being
+// used for all read and write operations. For production use this will typically be the
+// default namespace but for test usage this is set to a specific test namespace to both
+// avoid interfering with production data and to allow simpler cleanup and removal of
+// old temporary test data.
+//
+func (store *Store) SetNamespaceSuffix(suffix string) error {
+
```
Jim3Things, line 440: Can this be called more than once, and with different suffixes?  What happens if it is?
Also, how do we see the store getting cleaned out after running a unit test?

internal/clients/store/store_test.go
```diff
@@ -3,52 +3,100 @@
...
+		// Initialize() when using and ebmbedded Etcd store.
+		//
+		storeRoot.DefaultEndpoints = []string{fmt.Sprintf("%s:%v", cfg.Store.Test.EmbeddedHostname, cfg.Store.Test.EmbeddedPortClient)}
+
+		embedConfig := testEmbedConfig{
+			Node:       fmt.Sprintf("%s", cfg.Store.Test.EmbeddedHostname),
+			Addr:       fmt.Sprintf("%s", cfg.Store.Test.EmbeddedHostname),
+			PortClient: fmt.Sprintf("%v", cfg.Store.Test.EmbeddedPortClient),
+			PortPeer:   fmt.Sprintf("%v", cfg.Store.Test.EmbeddedPortPeer),
+			Path:       fmt.Sprintf("%s", cfg.Store.Test.EmbeddedPath),
```
Jim3Things, line 69: Why are the Sprintf("%s", ... cases using Sprintf?  Are we concerned about something around the source string?

internal/clients/store/store_test.go
```diff
@@ -148,26 +177,29 @@ func TestNew(t *testing.T) {
...
 	assert.Equal(t, getDefaultTimeoutConnect(), store.TimeoutConnect, "Mismatch in initialization of connection timeout")
 	assert.Equal(t, getDefaultTimeoutRequest(), store.TimeoutRequest, "Mismatch in initialization of request timeout")
 	assert.Equal(t, getDefaultTraceFlags(), store.TraceFlags, "Mismatch in initialization of trace flags")
+	assert.Equal(t, getDefaultNamespaceSuffix(), store.NamespaceSuffix, "Mismatch in initialization of namespace suffix")
 
 	endpoints := []string{"localhost:8080", "localhost:8181"}
 	timeoutConnect := getDefaultTimeoutConnect() * 2
 	timeoutRequest := getDefaultTimeoutRequest() * 3
 	traceFlags := traceFlagEnabled
+	namespaceSuffix := getDefaultNamespaceSuffix() + "/Suffix"
```
Jim3Things, line 193: I think this means that we can have one UT running at a time on a system.  Do we want to make the suffix random to allow for multiple simultaneous UT runs?  Or, also quite possible, did I read this wrong?

internal/config/settings.go
```diff
@@ -7,187 +7,273 @@
 package config
 
 import (
-    "context"
-    "fmt"
+	"context"
+	"fmt"
```
Jim3Things, line 11: fyi, there's a lot of tab/space changes here.  Not sure whether to do anything about that ...

internal/clients/store/store_test.go
MikeCJohnsonUk: No you read that right. That's what I was alluding to in the "specific namespace per test-instance" comment in the PR notes.

At present I'd claim this is unlikely, but it seems like it would be a good idea to "future proof" this to avoid test run mutual interference. 

How urgent?

internal/clients/store/store_test.go
MikeCJohnsonUk: I recall I was having an issue at one point, although it was likely to do with the port numbers as there was a confusion over whether these should be [uint, [u]int16 etc. 

In practice, I think having them as strings would be sufficient, since they are only used (I think) when building strings for URLs. If everything is a string then the Sprintf()'s should become redundent.

internal/clients/store/store.go
MikeCJohnsonUk: It can. Last writer wins. Since the store is not connected, it should be ok. Once connected, the namespace is fixed. 

I only really see the namespace suffix being used by the test code. CloudChamber itself is using only a single fixed namespace at present, but I have given that a version so it is is conceivable that there could be several, but I would not expect to see it change during a single invokation.

I'm on the fence about cleaning out after a run. One of the benefits of a well known namespace is that using the etcdctl tool it is easy to clean out an entire set of records with a single common key prefix (thats what a namespace effectively is). The question is, does the test have sufficient knowledge to be sure it's safe to clean out a test run dataset? Historically, I've never had a good answer to that. Clean it up, or delete it 'cos we're happy with the test outcome. Opinions? We could even make it a config setting ;-)

internal/config/settings.go
MikeCJohnsonUk: I keep on seeing these too. I've effectively given up on worry about it. It seems to change for no reason I can fathom, it does not seem to be consistent, it changes within the file, from edit to edit, from branch to branch, ...

internal/clients/store/store.go
Jim3Things: Can we clean it up at the start of a test run?  I most worry about past run data corrupting the next run's results.

internal/clients/store/store_test.go
Jim3Things: Ok

internal/clients/store/store_test.go
Jim3Things: I can't bring myself to say that it is urgent.  Not even sure if it is worth an issue.  Maybe just a comment here about how it assumes no multiple simultaneous UT runs, or somesuch...

internal/clients/store/store.go
MikeCJohnsonUk: Granted. However, if we have per-run uniquification, then there won't be an conflict, and also what counts as the previous run? If there is a fixed prefix then cleanup is easy(-ish) but we lose the ability to have parallel runs. 

Preference?

MikeCJohnsonUk: I've added an issue to add pre-test cleanup.

internal/clients/store/store_test.go
MikeCJohnsonUk: Actually, this is all going to be removed in the change for step 2 anyway. So I'm going to ignore it for now.

PR #59, opened 2020-06-12
Add new doc to help running and using etcd and etcdctl

MikeCJohnsonUk:
Title says it all.

### DIFF
--- a/docs/OperationalEnvironmentForWindows.md
+++ b/docs/OperationalEnvironmentForWindows.md
@@ -1,0 +1,156 @@
+At present the only requirement for external (to Cloud Chamber)
+components that need to be running is an etcd persistent store. This can
+be placed wherever is convenient but typically the local node is used.
+The shipping configuration file assumes that the etcd instance is
+running locally on the default ports. If this is not accurate, the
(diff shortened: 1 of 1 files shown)


docs/OperationalEnvironmentForWindows.md
```diff
@@ -0,0 +1,156 @@
...
+running locally on the default ports. If this is not accurate, the
+configuration file(s) need to be modified to suit.
+
+# Etcd Store
+
+The primary etcd documentation can be found at <https://etcd.io/docs/>
+with the most recent version (v3.4.0) being used.
+
+## Installing Etcd
+
```
Jim3Things, line 14: This lists a couple of ways to get etcd.  which one should we use when building cloud chamber?

MikeCJohnsonUk: Hmm. Actually it's either/or. But thinking about it, the startetcd.cmd script refers to %GOPATH% which effectively mean using "go get". I need to re-think this a little.

MikeCJohnsonUk:
This was superceeded in a different update.

PR #62, opened 2020-06-13
Remove protoc-gen-validate dependency

Jim3Things:
This change has three basic parts:
- proto files that referenced the protoc-gen-validate package have been edited to remove that dependency.

- Validations are now coded by hand in *.Validate.go files, which allow the bulk of the code to work unchanged.

- And there's script cleanup:
  - .gitignore is changed to retain the *.Validate.go files
  - buildall removes the qualifiers to invoke protoc-gen-validate during build
  - updateall, etc removes the the protoc-gen-validate package

There is an unrelated changes, except that it kept bugging me.  I've modified test.cmd to:
- take a trailing qualifier.  It defaults to nothing, which runs the tests in quiet mode, or you can set '-v' to get verbose tracing output.
- you can now specify either a single test suite name, or *, which runs all the tests.

So, 'test.cmd * -v' is the same as testall.cmd.  testall.cmd and testall.sh are now deleted.

Finally, once you have pulled this change, you should delete the directory subtree rooted at
%GOPATH%\src\github.com\envoyproxy\protoc-gen-validate, and then do a buildall.


pkg/protos/common/capacity.Validate.go
```diff
@@ -0,0 +1,50 @@
...
+)
+
+// Validate is a function to ensure that the blade capacity values are legal.
+// Note that since BladeCapacity is always used as a subfield, the Validate
+// function takes a prefix to use on the field name in order to place the
+// error correctly
+func (x *BladeCapacity) Validate(prefix string) error {
+    // A blade must have at least one core
+    if x.Cores <= 0 {
+        return ErrMustBeGTE{
```
MikeCJohnsonUk, line 15: Should this error be GT not GTE?

pkg/protos/common/capacity.Validate.go
```diff
@@ -0,0 +1,50 @@
...
+            Field:    fmt.Sprintf("%sCores", prefix),
+            Actual:   x.Cores,
+            Required: 1,
+        }
+    }
+
+    // .. and it must have some memory
+    if x.MemoryInMb <= 0 {
+        return ErrMustBeGTE{
```
MikeCJohnsonUk, line 24: Should this error be GT not GTE?

pkg/protos/common/capacity.Validate.go
```diff
@@ -0,0 +1,50 @@
...
+            Field:    fmt.Sprintf("%sMemoryInMb", prefix),
+            Actual:   x.MemoryInMb,
+            Required: 1,
+        }
+    }
+
+    // .. and some disk space
+    if x.DiskInGb <= 0 {
+        return ErrMustBeGTE{
```
MikeCJohnsonUk, line 33: Should this error be GT not GTE?

pkg/protos/common/capacity.Validate.go
```diff
@@ -0,0 +1,50 @@
...
+            Field:    fmt.Sprintf("%sDiskInGb", prefix),
+            Actual:   x.DiskInGb,
+            Required: 1,
+        }
+    }
+
+    // .. and a network bandwidth allowance
+    if x.NetworkBandwidthInMbps <= 0 {
+        return ErrMustBeGTE{
```
MikeCJohnsonUk, line 42: Should this error be GT not GTE?

pkg/protos/common/errors.go
```diff
@@ -0,0 +1,54 @@
...
+    Actual int64
+    Required int64
+}
+
+func (e ErrMustBeGTE) Error() string {
+    return fmt.Sprintf(
+        "the field %q must be greater than or equal to %d.  It is %d, which is invalid",
+        e.Field,
+        e.Required,
+        e.Actual)
```
MikeCJohnsonUk, line 21: Given the tests in capacity.Validate.go we should probably also have ErrMustBeGT

pkg/protos/Stepper/Stepper.Validate.go
```diff
@@ -0,0 +1,61 @@
...
+import (
+    "github.com/Jim3Things/CloudChamber/pkg/protos/common"
+)
+
+// Validate is a function to ensure that the policy request values are
+// legal
+func (x *PolicyRequest) Validate() error {
+    // The policy must be within the valid range for the enum
+    p := int64(x.Policy)
+    if p > 0 && p < 4 {
```
MikeCJohnsonUk, line 13: I don't suppose there's any way the '1' and '4' can be symbolized is there?

pkg/protos/common/capacity.Validate.go
Jim3Things: The required sets the minimum, so it is GTE wrt the required of 1.  I could make the tests <1, just to be consistent

pkg/protos/Stepper/Stepper.Validate.go
Jim3Things: I'll turn this into a switch statement, which is a little less obtuse.

pkg/protos/common/capacity.Validate.go
MikeCJohnsonUk: either GT 0 or GTE 1, dealers choice!

pkg/protos/Stepper/Stepper.Validate.go
MikeCJohnsonUk: That gets a "Like" using the modern vernacular.

PR #64, opened 2020-06-21
Add stepper REST head, UT, trace channel sync

Jim3Things:
This adds REST processing for GET /api/stepper/now, and a UT to verify
In turn, this caused:
- changes to timestamp setup,
- adding a default stepper policy to the config
- Tracking down and fixing a trace/test context sync issue

### DIFF
--- a/internal/services/frontend/frontend.go
+++ b/internal/services/frontend/frontend.go
@@ -29,6 +29,7 @@ import (
 
 	"github.com/Jim3Things/CloudChamber/internal/config"
 	st "github.com/Jim3Things/CloudChamber/internal/tracing/server"
+	pb "github.com/Jim3Things/CloudChamber/pkg/protos/admin"
 )
 
 // Computer is a representation an individual Computer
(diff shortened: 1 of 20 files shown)


internal/services/frontend/frontend.go
```diff
@@ -16,43 +16,21 @@
...
 	"net/http"
 	"strings"
-	"sync"
 
 	"github.com/gorilla/mux"
 	"github.com/gorilla/securecookie"
 	"github.com/gorilla/sessions"
+	"google.golang.org/grpc"
 
+	clients "github.com/Jim3Things/CloudChamber/internal/clients/timestamp"
```
MikeCJohnsonUk, line 29: Probably picky, but this import used an alias of "ts" in common_test.go - any chance we could do the same?

internal/tracing/exporters/unit_test/utIntegration.go
```diff
@@ -20,17 +20,46 @@ type Exporter struct {
 
 var testContext *testing.T
 
+// Since the actor system can produce async activity that occurs outside of an
+// active test context, we have to be able to handle these events.  When they
+// happen we save them into this array and process them as soon as we see an
+// active test context.
+var savedEntries []*log.Entry
+
```
MikeCJohnsonUk, line 28: A though - do we tag the - re: testContext activity in any way to indicate it happened "between" active tests? Is that even interesting, assuming it's do-able?

Jim3Things: Good idea.  I've added '(deferred)' to the span line in that case

internal/services/frontend/frontend.go
Jim3Things: Agree. Done.

PR #65, opened 2020-06-25
Rename inventory test file

Jim3Things:
In one of our discussions I agreed to rename the inventory test tile to better match what it was testing.  This seemed like an opportune time to do that.

Also include two minor comment formatting cleanups.


Jim3Things:
This was pulled from the master branch too high up.

PR #71, opened 2020-06-26
Update store and store test tracing to use current trace methods

MikeCJohnsonUk:
Update store and store test tracing to use current trace methods to avoid special handling during test initialization when calling tracing routines from TestMain() (specifically ReadGlobalConfig())

Also cleans up some trace message to remove spurious "\n" characters, attempts to cleanup processing tracing of store call responses containing multiple values and a few other cleanup issues

### DIFF
--- a/internal/clients/store/store.go
+++ b/internal/clients/store/store.go
@@ -102,6 +102,9 @@ type TraceFlags uint
 const (
 	traceFlagEnabled TraceFlags = 1 << iota
 	traceFlagExpandResults
+	traceFlagTraceKey
+	traceFlagTraceKeyAndValue
+	traceFlagExpandResultsInTest
 )
 
 type global struct {
@@ -309,7 +312,7 @@ func (store *Store) logEtcdResponseError(ctx context.Context, err error) {
 			st.Infof(ctx, -1, "ctx is attached with a deadline is exceeded: %v", err)
 
 		case rpctypes.ErrEmptyKey:
-			st.Infof(ctx, -1, "client-side error: %v\n", err)
+			st.Infof(ctx, -1, "client-side error: %v", err)
 
 		default:
 			if ev, ok := status.FromError(err); ok {
@@ -560,12 +563,12 @@ func (store *Store) GetClusterMembers() (result *Cluster, err error) {
 
 			if store.trace(traceFlagExpandResults) {
 				for i, node := range result.Members {
-					st.Infof(ctx, -1, "node [%v] Id: %v Name: %v\n", i, node.ID, node.Name)
+					st.Infof(ctx, -1, "node [%v] Id: %v Name: %v", i, node.ID, node.Name)
 					for i, url := range node.ClientURLs {
-						st.Infof(ctx, -1, "  client [%v] URL: %v\n", i, url)
+						st.Infof(ctx, -1, "  client [%v] URL: %v", i, url)
 					}
 					for i, url := range node.PeerURLs {
-						st.Infof(ctx, -1, "  peer [%v] URL: %v\n", i, url)
+						st.Infof(ctx, -1, "  peer [%v] URL: %v", i, url)
 					}
 				}
 			}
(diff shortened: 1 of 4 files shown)


internal/clients/store/store.go
```diff
@@ -264,11 +221,24 @@ func (store *Store) disconnected(op string) error {
...
+		storeRoot.DefaultEndpoints = []string{fmt.Sprintf("%s:%v", cfg.Store.EtcdService.Hostname, cfg.Store.EtcdService.Port)}
+		storeRoot.DefaultTimeoutConnect = time.Duration(cfg.Store.ConnectTimeout) * time.Second
+		storeRoot.DefaultTimeoutRequest = time.Duration(cfg.Store.RequestTimeout) * time.Second
+		storeRoot.DefaultTraceFlags = TraceFlags(cfg.Store.TraceLevel)
+		storeRoot.DefaultNamespaceSuffix = ""
+
+		st.Infof(
+			ctx,
+			-1,
+			"    EP: %v TimeoutConnect: %v TimeoutRequest: %v DefTrcFlags: %v NsSuffix: %v",
```
Jim3Things, line 234: Why the leading spaces here?

internal/clients/store/store.go
```diff
@@ -349,10 +321,10 @@ func (store *Store) logEtcdResponseError(err error) {
 					// server-side context might have timed-out first (due to clock skew)
 					// while original client-side context is not timed-out yet
 					//
-					log.Printf("server-side deadline is exceeded: %v\n", code)
+					st.Infof(ctx, -1, "server-side deadline is exceeded: %v", code)
 				}
 			} else {
-				log.Printf("bad cluster endpoints, which are not etcd servers: %v\n", err)
+				st.Infof(ctx, -1, "bad cluster endpoints, which are not etcd servers: %v", err)
```
Jim3Things, line 327: Do these all make sense as info traces?  Should some be error traces?

internal/clients/store/store.go
```diff
@@ -380,12 +355,16 @@ func (store *Store) GetTraceFlags() TraceFlags { return store.TraceFlags }
 //
 func (store *Store) SetAddress(endpoints []string) (err error) {
 
-	if err := store.connected("Set(address)"); err != nil {
-		return err
-	}
+	return st.WithSpan(context.Background(), tracing.MethodName(1), func(ctx context.Context) (err error) {
+		if err := store.connected(ctx, "SetAddress"); err != nil {
```
Jim3Things, line 359: There's something I'm really torn on with the pattern here.  It looks like op string on the store.connected call is pretty much always the name of the method calling store.connected.  I keep wondering if you could use tracing.MethodName(n), or some close variant rather than pass it in.  But I'm not sure if the savings in typing is overbalanced by obscurity in the result.  But I wanted to at least pass the comment along.

internal/clients/store/store.go
```diff
@@ -803,99 +833,106 @@ func (store *Store) ReadWithPrefix(keyPrefix string) (result []KeyValueResponse,
...
 
-	return err
+		return err
+	})
 }
 
 // DeleteMultiple is a method that can be used to remove a set of key/value pairs.
 //
 // This is essentially a convenience method to allow multiuple values to be fetched
 // in a single call rather than repeating individual calls to the Delete() method.
```
Jim3Things, line 864: nit: fetched -> removed?

internal/clients/store/store.go
MikeCJohnsonUk: Yes. Obvious in hindsight. Sigh...

internal/clients/store/store.go
MikeCJohnsonUk: Well, the reason for the op string in the first placed was to allow the trace of the reason for the connect attempt in the event of a failure. Now that tracing is happening in the context of a span, which already has a name of the calling func, I don't believe the op string is required at all.

The same is true of store.disconnected.

So, I'll remove the op string altogether as being redundant.

internal/clients/store/store.go
MikeCJohnsonUk: Removed.

internal/clients/store/store.go
MikeCJohnsonUk: Agreed.

PR #72, opened 2020-06-27
HandlerBladeList

manpreet85:
In this attempt I have written code to scan the list of blades in a rack. Please review and help me finetune if I am in correct direction. Cheers, Manpreet Singh @Jim3Things @MikeCJohnsonUk


internal/services/frontend/DBInventory.go
```diff
@@ -81,3 +81,39 @@ func (m *DBInventory) Get(rackid string) (*pb.ExternalRack, error) {
...
+	}
+	for name := range r.Blades {
+
+		if err := action(name); err != nil {
+			return err
+		}
+
+	}
+	return nil
+}
```
Jim3Things, line 102: The logic here is fine, but there is some tidying to do with the code.  There should be a comment just before the function that describes is (and it should start "// ScanBladesInRack is " and then a description of what it does.

White space (blank lines) also matter in terms of grouping logical thoughts together.  Personally, I'd add one between 93 & 94 (getting the rack is one thought, looping through the blades is a second.  I'd remove the ones at 95 & 99, as looping through the blades is one thought only.  I'd also add one between 100 & 101, as returning all done is the final thought in the function.

My personal choice would also be to remove the one at 86, but that is only personal taste.  Mike, for instance, does it the same way you do.

internal/services/frontend/DBInventory.go
```diff
@@ -81,3 +81,39 @@ func (m *DBInventory) Get(rackid string) (*pb.ExternalRack, error) {
...
+// GET api/racks/rack1/blades  ## 91 and 93 as 1st check in
+
+//Get blade list from rack 1 <-- uris for blade1 & blade2 in rack1
+
+// GET api/racks/rack1/blades/1  ## 95 and 97 as second checkin
+
+// <-- details for blade 1 in rack1
+
+// RAck 2 as well
+
```
Jim3Things, line 118: These are fine for now as notes to yourself, but they will have to be removed for the final checkin

internal/services/frontend/inventory.go
```diff
@@ -48,6 +48,7 @@ func inventoryAddRoutes(routeBase *mux.Router) {
 	//
 	//routeRacks.HandleFunc("/racks/{rackid}", handlerRacksCreate).Methods("POST", "GET") // May be only GET
 	routeRacks.HandleFunc("/{rackid}", handlerRacksRead).Methods("GET")
+	routeRacks.HandleFunc("/{rackid}/blades")
```
Jim3Things, line 51: This is missing some important parts.  The line needs to say what path pattern it is processing, what function to call when a path is supplied that matches the pattern, and which http methods this applies to.  The path can include segments (the text between two adjacent '/' characters) that should be treated as variables that the function that handles this path may be able to call and use.

The line above (50) says, for example, that any path that gets to this part of the route (which we know from the earlier processing is anything that has '/api/racks' as a prefix, has a final segment, and using the http verb 'GET', should call handlerRacksRead.  Furthermore, that last segment is available to handlerRacksRead as a variable ("rackid") that it can fetch.

Line 51 is missing what handler to call and what http verbs it applies to.  You need to add those.

internal/services/frontend/inventory.go
```diff
@@ -115,6 +116,35 @@ func handlerRacksRead(w http.ResponseWriter, r *http.Request) {
 	})
 }
 
+func handlerBladeList(w http.ResponseWriter, r http.Request) {
+	_ = st.WithSpan(context.Background(), tracing.MethodName(1), func(ctx context.Context) (err error){
+		vars := mux.Vars(r)
+		rackid := vars["rackid"] // captured the key value in rackid variable
+	
+		if _, err := fmt.Fprintln(w, "Racks (List)"); err != nil { // Why printing the rack list in here?
+			return httpError(ctx, w, err)
```
Jim3Things, line 125: You certainly want to change the title.  Since the title should be informative, I'd think it should also include the rack name it applies to.  So we'll want to use fmt.Fprintf, rather than fmt.Fprintln - this lets us put a format in place and add values that get put into the string.

So, if we assume we'll want to put out, say, "Blades in rack1 (List)\n" as the title line, then we'd replace fmt.Fprintln(w, "Racks (List)") with fmt.Fprintf(w, "Blades in %s (List)\n", rackid)

If you want to put out a different title line than that, that's fine - use fmt.Fprintf to format it.

internal/services/frontend/inventory.go
```diff
@@ -115,6 +116,35 @@ func handlerRacksRead(w http.ResponseWriter, r *http.Request) {
...
+		}
+
+		b := r.URL.String()
+		if !strings.HasSuffix(b, "/") {
+			b += "/"
+		}
+
+		return dbInventory.ScanBladesinRack(rackid, func(bladeid int64) (err error) ) {
+
+			target := fmt.Sprintf("%s%s", b, bladeid)
```
Jim3Things, line 135: This has an issue with the format string.  It uses %s%s, which means that it expects b is a string (which it is) and bladeid is a string (which it isn't).  The correct format would be %s%d, which means that it expects b is a string, and bladeid is an integer.

If you're interested, the format string argument options are described in https://golangdocs.com/string-formatting-in-golang and, more completely, in https://golang.org/pkg/fmt/

internal/services/frontend/inventory.go
```diff
@@ -115,6 +116,35 @@ func handlerRacksRead(w http.ResponseWriter, r *http.Request) {
...
+
+			st.Infof(ctx, -1, " Listing blades '%s' at '%s'", bladeid, target)
+			if _, err = fmt.Fprintln(w, target); err != nil {
+				return httpError(ctx, w, err)
+			}
+			return nil
+		})  
+
+	})
```
Jim3Things, line 144: As with the ScanBladesinRack, you should look over the blank lines to see if you're grouping thoughts together properly.

internal/services/frontend/DBInventory.go
```diff
@@ -81,3 +81,39 @@ func (m *DBInventory) Get(rackid string) (*pb.ExternalRack, error) {
 	return r, nil
 
 }
+
+func (m *DBInventory) ScanBladesinRack(rackid string, action func(bladeid int64) error) error {
+
```
Jim3Things, line 86: The UpAndDown naming style should have an Up for every word, so this should be ScanBladesInRack.

internal/services/frontend/inventory.go
Jim3Things: There are two functions you'll want, depending on the path.  The basic idea for the path structure is that there are paths that hold collections, such as /api/users holds the list of all users, /api/racks holds the list of all racks.  And /api/racks/{rackid}/blades holds the list of all blades in the rack specified by {rackid}.  That path will want a handleBladesList function.

The second pattern is a path that ends with a variable, and that sort of path is involved with the details of that specific entry.  So /api/users/admin involves the details of the user 'admin', /api/racks/{rackid} involves the details of the rack specified by {rackid}.  And /api/racks/{rackid}/blades/{bladeid} involves the details of the blade specified by {bladeid} in the rack specified by {rackid}.  That path will want a handleBladeRead function.

internal/services/frontend/inventory.go
```diff
@@ -48,7 +48,7 @@ func inventoryAddRoutes(routeBase *mux.Router) {
 	//
 	//routeRacks.HandleFunc("/racks/{rackid}", handlerRacksCreate).Methods("POST", "GET") // May be only GET
 	routeRacks.HandleFunc("/{rackid}", handlerRacksRead).Methods("GET")
-	routeRacks.HandleFunc("/{rackid}/blades")
+	routeRacks.HandleFunc("/{rackid}/blades",)
```
Jim3Things, line 51: This will want a function handleBladesList, and a method of "GET"

internal/services/frontend/inventory.go
```diff
@@ -121,7 +121,7 @@ func handlerBladeList(w http.ResponseWriter, r http.Request) {
 		vars := mux.Vars(r)
 		rackid := vars["rackid"] // captured the key value in rackid variable
 	
-		if _, err := fmt.Fprintln(w, "Racks (List)"); err != nil { // Why printing the rack list in here?
+		if _, err := fmt.Fprintf(w, "Blades in %s (List)\n", rackid); err != nil { // Why printing the rack list in here?
```
Jim3Things, line 124: Remove the comment, as that question has been answered, I assume.

internal/services/frontend/inventory.go
```diff
@@ -130,20 +130,45 @@ func handlerBladeList(w http.ResponseWriter, r http.Request) {
...
 		}
 
-		return dbInventory.ScanBladesinRack(rackid, func(bladeid int64) (err error) ) {
+		return dbInventory.ScanBladesInRack(rackid, func(bladeid int64) (err error) ) {
 
-			target := fmt.Sprintf("%s%s", b, bladeid)
-
-			st.Infof(ctx, -1, " Listing blades '%s' at '%s'", bladeid, target)
+			target := fmt.Sprintf("%s%d", b, bladeid)
+						st.Infof(ctx, -1, " Listing blades '%d' at '%s'", bladeid, target)
```
Jim3Things, line 136: Why is this indented so far?

internal/services/frontend/inventory.go
```diff
@@ -130,20 +130,45 @@ func handlerBladeList(w http.ResponseWriter, r http.Request) {
...
+		}  
 
 	})
 }
 
+func handlerBladeRead(w http.ResponseWriter, r http.Request){
+	_ = st.WithSpan(context.Background(), tracing.MethodName(1), func(ctx context.Context) (err error){
+		vars := mux.Vars(r)
+		rackid := vars["rackid"] // captured the key value in rackid variable
+
```
Jim3Things, line 151: The bladeid should be another variable in the path (see my comment back on your question)

internal/services/frontend/inventory.go
```diff
@@ -130,20 +130,45 @@ func handlerBladeList(w http.ResponseWriter, r http.Request) {
...
+	
+		if _, err := fmt.Fprintf(w, "Blades in %s (List)\n", rackid); err != nil { // Why printing the rack list in here?
+			return httpError(ctx, w, err)
+		}
+
+		b := r.URL.String()
+		if !strings.HasSuffix(b, "/") {
+			b += "/"
+		}
+		return dbInventory.ScanBladesInRack(rackid, func(bladeid int64) (err error) ) {
```
Jim3Things, line 160: You shouldn't need to scan for the blade.  You know the rack id and the blade id, that should be enough to directly look it up.

internal/services/frontend/inventory.go
```diff
@@ -130,20 +130,45 @@ func handlerBladeList(w http.ResponseWriter, r http.Request) {
...
+		w.Header().Set("Content-Type", "application/json")
+
+		st.Infof(ctx, -1, "Returning details for blade %d: %v", bladeid, u)
+
+		// Get the user entry, and serialize it to json
+		// (export userPublic to json and return that as the body)
+		p := jsonpb.Marshaler{}
+		return p.Marshal(w, u)
```
Jim3Things, line 168: You'll want to update the comment.  I'd also suggest that u might not be a great variable name for a blade details entry once you've retrieved it.  Think about a different variable name.

internal/services/frontend/inventory.go
manpreet85: and PUT as well? we would want to add blades as well.

manpreet85:
Hi Jim,

I am thinking the following as a test when encounter “String” as a blade. I am not sure how to execute it though. Can you please help?
func TestInventoryStringBlades(t *testing.T) {
    unit_test.SetTesting(t)
    defer unit_test.SetTesting(nil)

    request := httptest.NewRequest("GET", fmt.Sprintf("%s%s", baseURI, "/api/racks/rack1/blades/Jeff"), nil)
    request.Header.Set("Content-Type", "application/json")
    response := doHTTP(request, nil)
    actualObj, err := 
    if assert.Errorf(t , err , "error message %s", "formatted"){
    assert.Equal(t, response, err)
    }
}

internal/services/frontend/inventory.go
```diff
@@ -116,58 +117,57 @@ func handlerRacksRead(w http.ResponseWriter, r *http.Request) {
...
-func handlerBladeList(w http.ResponseWriter, r http.Request) {
-	_ = st.WithSpan(context.Background(), tracing.MethodName(1), func(ctx context.Context) (err error){
+func handlerBladeList(w http.ResponseWriter, r *http.Request) {
+	_ = st.WithSpan(context.Background(), tracing.MethodName(1), func(ctx context.Context) (err error) {
 		vars := mux.Vars(r)
 		rackid := vars["rackid"] // captured the key value in rackid variable
-	
-		if _, err := fmt.Fprintf(w, "Blades in %s (List)\n", rackid); err != nil { // Why printing the rack list in here?
+
+		if _, err := fmt.Fprintf(w, "Blades in %s (List)\n", rackid); err != nil {
```
Jim3Things, line 125: It's a nit, and sorry that I didn't spot it earlier, but I'd use %q rather than %s for the rackid.  %s is a directive to put the string contents into the formatted output as written.  %q puts a quoted form of the string into the output.  That means that things like trailing spaces don't get lost.

internal/services/frontend/inventory.go
```diff
@@ -116,58 +117,57 @@ func handlerRacksRead(w http.ResponseWriter, r *http.Request) {
...
 			b += "/"
 		}
-
-		return dbInventory.ScanBladesInRack(rackid, func(bladeid int64) (err error) ) {
+		return dbInventory.ScanBladesInRack(rackid, func(bladeid int64) error {
 
 			target := fmt.Sprintf("%s%d", b, bladeid)
-						st.Infof(ctx, -1, " Listing blades '%d' at '%s'", bladeid, target)
-			
+			st.Infof(ctx, -1, " Listing blades '%d' at '%s'", bladeid, target)
```
Jim3Things, line 135: Following on from the comment above, I think it would be useful to replace the '%s' with %q in the format string above (and anywhere else it appears in inventory.go.  It helps with consistency.

internal/services/frontend/inventory_test.go
```diff
@@ -70,3 +72,63 @@ func TestInventoryUnknownRack(t *testing.T) {
...
+	var splits = strings.Split(s, "\n") // Created an array per line
+
+	expected := []string{"/api/racks/rack1/blades/1", "/api/racks/rack1/blades/2", ""}
+
+	assert.Equal(t, splits[0], "Blades in rack1 (List)")
+	assert.ElementsMatch(t, expected, splits[1:])
+
+	assert.Nil(t, err)
+}
+func TestInventoryUnknownBlades(t *testing.T) {
```
Jim3Things, line 95: There are two comments on this test and the others below.  Both are just to help with clarity.  First, any read test should use the singular rather than the plural form for the object under test, and any list test should use the plural.  This reflects that List operations work against a set, and Read operations work against just one object.

So we have TestInventoryListBlades, which is good.  But I'd recommend changing this test name to TestInventoryUnknownBlade.

internal/services/frontend/inventory_test.go
```diff
@@ -70,3 +72,63 @@ func TestInventoryUnknownRack(t *testing.T) {
...
+}
+func TestInventoryStringBlades(t *testing.T) {
+	unit_test.SetTesting(t)
+	defer unit_test.SetTesting(nil)
+
+	request := httptest.NewRequest("GET", fmt.Sprintf("%s%s", baseURI, "/api/racks/rack1/blades/Jeff"), nil)
+	request.Header.Set("Content-Type", "application/json")
+	response := doHTTP(request, nil)
+
+	assert.Equal(t, http.StatusNotFound, response.StatusCode, "Handler returned the expected error: %v", response.StatusCode)
```
Jim3Things, line 133: I think this is going to have a different error - it should be http.StatusBadRequest (based on line 157 in inventory.go)

Jim3Things:
Hi Manpreet,

The test you wrote in the commit that I just reviewed is mostly what you
want here.  I don't think it needs to be more extensive.

If you do want to expand on it, I think you're trying here to get the error
text and test that.  If so, then you want the body as an array of bytes,
and then to convert it into a string for validation.

To get the body as an array of bytes, use the getBody function, as in:

body, err := getBody(response)
assert.Nilf(t, err, "unexpected error while getting the body, err=%v", err)

Then to convert it into a string, you use the string function, as in:

s := string(body)

then you just want a final assert that verifies the string you're looking
for, as in:

assert.Equal(t, <expected string>, s)

And, just like we did with the blade list header, I'd put a simple bogus
string in place of <expected string>, run the test, and use that to figure
out what the <expected string> should be.

HTH,
Jim.

manpreet85:
Tests passed and created new tests for blades @Jim3Things @MikeCJohnsonUk

PR #75, opened 2020-07-10
Session management updates

Jim3Things:
Add session management to stepper frontend handlers
Add session manager to handle session state
Move users.go to use the session manager

### DIFF
--- a/internal/services/frontend/session_manager.go
+++ b/internal/services/frontend/session_manager.go
@@ -3,17 +3,169 @@ package frontend
 
 import (
     "context"
+    "fmt"
     "net/http"
+    "sync"
+    "time"
 
     "github.com/gorilla/sessions"
 
     pb "github.com/Jim3Things/CloudChamber/pkg/protos/admin"
 )
 
+const (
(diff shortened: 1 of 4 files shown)


internal/services/frontend/session_manager.go
```diff
@@ -3,25 +3,177 @@ package frontend
...
+// getSession is a function that returns the state associated with the current
+// session.  It also returns a true/false flag indicating if the state was
+// found in the active sessions, much like map lookup does.
+func getSession(session *sessions.Session) (SessionState, bool) {
+    mutex.Lock()
+    defer mutex.Unlock()
+
+    purgeStaleSessions()
+
+    if id, ok := session.Values[sessionIdKey].(int64); ok {
```
MikeCJohnsonUk, line 93: Ok for my education, I gotta ask, what does the ".(int64)" part do? Looks like a cast of some form, but...?

Jim3Things: it's a type cast.  Values is an object, .(int64) casts it to an int64.  If it isn't that, ok is set to false.

PR #76, opened 2020-07-11
Update startetcd script to allow for system path search

MikeCJohnsonUk:
Currently the script assumes the etcd.exe is always located in the %GOPATH%\bin directory. This works well when using an etcd.exe built from source, but is not always convenient when using pre-built binaries installed elsewhere.

The update allows the binaries to either be in a directory identified with the ETCDBINPATH environment variable, or in any directory already on the system path. 

This is in addition to the current %GOPATH%\bin directory.


scripts/startetcd.cmd
```diff
@@ -37,6 +37,28 @@ if /i "%1" == "-?" (goto :StartEtcdHelp)
 if /i "%1" == "-h" (goto :StartEtcdHelp)
 
 
+rem Find a binary to use
+rem
```
Jim3Things, line 41: Should there be a shell version as well?

MikeCJohnsonUk: Probably yes. Basically we should really have shell versions of everything.

PR #78, opened 2020-07-11
Add transactional read and update support to the store layer and add user record CRUD operations

MikeCJohnsonUk:
The aim of this change is the first cut of actually storing useful data in the underlying store, in particular the "user" records as already defined. This is done using the transactional (txn) support provided by etcd which allows for conditional updates to the store based upon the current revisions of the store and the revision of the record being modified.

At this point the user record layer is a little simplistic and will be simplified using a more generic style if interface in a future update.

Any suggestions for trace improvements would be welcome.

Testing of conditions other than unconditional and equal will be added in subsequent releases as required.

### DIFF
--- a/internal/clients/store/common_test.go
+++ b/internal/clients/store/common_test.go
@@ -69,7 +69,6 @@ func commonSetup() {
 	}
 
 	setDefaultNamespaceSuffix(testNamespace)
-	return
 }
 
 func cleanNamespace(testNamespace string) error {
--- a/internal/clients/store/common_test.go
+++ b/internal/clients/store/common_test.go
@@ -130,6 +130,26 @@ func testGenerateKeySetFromKeyValueSet(keyValueSet []KeyValueArg) []string {
 	return keySet
 }
 
+// Build a set of key,value pairs to be created unconditionally in the store.
+//
+func testGenerateRecordUpdateSetFromKeyValueSet(keyValueSet []KeyValueArg, label string, condition WriteCondition) RecordUpdateSet {
+
+	recordUpdateSet := RecordUpdateSet{Label: label, Records: make(map[string]RecordUpdate)}
+
+	for _, kv := range keyValueSet {
+		recordUpdateSet.Records[kv.key] =
+			RecordUpdate{
+				Condition: condition,
+				Record: Record{
+					Revision: RevisionInvalid,
+					Value:    kv.value,
+				},
+			}
+	}
+
+	return recordUpdateSet
+}
+
 // TestMain is the Common test startup method.  This is the _only_ Test* function in this
 // file.
 func TestMain(m *testing.M) {
--- a/internal/clients/store/common_test.go
+++ b/internal/clients/store/common_test.go
@@ -1,0 +1,138 @@
+package store
+
+import (
+	"flag"
+	"fmt"
+	"log"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/Jim3Things/CloudChamber/internal/config"
+	"github.com/Jim3Things/CloudChamber/internal/tracing/exporters"
+	"github.com/Jim3Things/CloudChamber/internal/tracing/setup"
+)
+
+// A number of tests use a pre-computed set of keys for the purposes of
+// the test. This constant determines the standard size of these sets.
+// The value chosen is largely arbitrary. Selecting a value that is too
+// large may lead to problems with values to /from the underlying store,
+// so be reasonable.
+//
+const keySetSize = 100
+
+var (
+	baseURI     string
+	initialized bool
+
+	testNamespaceSuffixRoot = "/Test"
+
+	configPath *string
+)
+
+func commonSetup() {
+	var testNamespace string
+
+	setup.Init(exporters.UnitTest)
+
+	configPath = flag.String("config", ".", "path to the configuration file")
+	flag.Parse()
+
+	cfg, err := config.ReadGlobalConfig(*configPath)
+	if err != nil {
+		log.Fatalf("failed to process the global configuration: %v", err)
+	}
+
+	Initialize(cfg)
+
+	// It is meaningless to have both a unique per-instance test namespace
+	// and to clean the store before the tests are run
+	//
+	if cfg.Store.Test.UseUniqueInstance && cfg.Store.Test.PreCleanStore {
+		log.Fatalf("invalid configuration: both UseUniqueInstance and PreCleanStore are enabled: %v", err)
+	}
+
+	// For test purposes, need to set an alternate namespace rather than
+	// rely on the standard. From the configuration, we can either use the
+	// standard, fixed, well-known prefix, or we can use a per-instance
+	// unique prefix derived from the current time
+	//
+	if cfg.Store.Test.UseUniqueInstance {
+		testNamespace = fmt.Sprintf("%s/%s/", testNamespaceSuffixRoot, time.Now().Format(time.RFC3339Nano))
+	} else {
+		testNamespace = testNamespaceSuffixRoot + "/Standard/"
+	}
+
+	if cfg.Store.Test.PreCleanStore {
+		if err := cleanNamespace(testNamespace); err != nil {
+			log.Fatalf("failed to pre-clean the store as requested - namespace: %s err: %v", testNamespace, err)
+		}
+	}
+
+	setDefaultNamespaceSuffix(testNamespace)
+	return
+}
+
+func cleanNamespace(testNamespace string) error {
+
+	store := NewStore()
+
+	if store == nil {
+		log.Fatalf("unable to allocate store context for pre-cleanup")
+	}
+
+	if err := store.SetNamespaceSuffix(""); err != nil {
+		return err
+	}
+	if err := store.Connect(); err != nil {
+		return err
+	}
+	if err := store.DeleteWithPrefix(testNamespace); err != nil {
+		return err
+	}
+
+	store.Disconnect()
+
+	return nil
+}
+
+func testGenerateKeyValueSet(setSize int, setName string) []KeyValueArg {
+
+	keyValueSet := make([]KeyValueArg, setSize)
+
+	for i := range keyValueSet {
+		keyValueSet[i].key = fmt.Sprintf("%s/Key%04d", setName, i)
+		keyValueSet[i].value = fmt.Sprintf("%s/Val%04d", setName, i)
+	}
+
+	return keyValueSet
+}
+
+func testGenerateKeyValueMapFromKeyValueSet(keyValueSet []KeyValueArg) map[string]string {
+
+	keyValueMap := make(map[string]string, len(keyValueSet))
+
+	for _, kv := range keyValueSet {
+		keyValueMap[kv.key] = kv.value
+	}
+
+	return keyValueMap
+}
+
+func testGenerateKeySetFromKeyValueSet(keyValueSet []KeyValueArg) []string {
+
+	keySet := make([]string, len(keyValueSet))
+
+	for i, kv := range keyValueSet {
+		keySet[i] = kv.key
+	}
+
+	return keySet
+}
+
+// TestMain is the Common test startup method.  This is the _only_ Test* function in this
+// file.
+func TestMain(m *testing.M) {
+	commonSetup()
+	os.Exit(m.Run())
+}
--- a/internal/clients/store/common_test.go
+++ b/internal/clients/store/common_test.go
@@ -22,7 +22,6 @@ import (
 const keySetSize = 100
 
 var (
-	baseURI     string
 	initialized bool
 
 	testNamespaceSuffixRoot = "/Test"
@@ -132,7 +131,7 @@ func testGenerateKeySetFromKeyValueSet(keyValueSet []KeyValueArg) []string {
 
 // Build a set of key,value pairs to be created unconditionally in the store.
 //
-func testGenerateRecordUpdateSetFromKeyValueSet(keyValueSet []KeyValueArg, label string, condition WriteCondition) RecordUpdateSet {
+func testGenerateRecordUpdateSetFromKeyValueSet(keyValueSet []KeyValueArg, label string, condition Condition) RecordUpdateSet {
 
 	recordUpdateSet := RecordUpdateSet{Label: label, Records: make(map[string]RecordUpdate)}
 
--- a/internal/clients/store/errors.go
+++ b/internal/clients/store/errors.go
@@ -25,19 +25,6 @@ var (
 	ErrStoreConnected = errors.New("CloudChamber: client currently connected")
 )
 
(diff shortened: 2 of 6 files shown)


internal/clients/store/storeapi.go
```diff
@@ -0,0 +1,324 @@
...
+		if readResponse, err = store.ReadMultipleTxn(ctx, recordKeySet); err != nil {
+			return err
+		}
+
+		recordCount := len(readResponse.Records)
+
+		switch recordCount {
+		default:
+			st.Errorf(ctx, -1, "searching for user %q found %v records when expecting just one", name, recordCount)
+			return ErrStoreBadRecordCount(name)
```
Jim3Things, line 220: Why default at the top in this case?

internal/clients/store/storeapi.go
```diff
@@ -0,0 +1,324 @@
...
+			userName := strings.TrimPrefix(k, storeRootUsers)
+
+			if userName != u.Name {
+				return ErrStoreBadRecordContent(k)
+			}
+
+			rs.Records[userName] = UserRecord{Revision: r.Revision, User: u}
+
+			st.Infof(ctx, -1, "found record for user %q eith revision %v", u.Name, r.Revision)
+		}
```
Jim3Things, line 312: Nit: typo - eith should be with

internal/clients/store/common_test.go
```diff
@@ -0,0 +1,158 @@
...
+// A number of tests use a pre-computed set of keys for the purposes of
+// the test. This constant determines the standard size of these sets.
+// The value chosen is largely arbitrary. Selecting a value that is too
+// large may lead to problems with values to /from the underlying store,
+// so be reasonable.
+//
+const keySetSize = 100
+
+var (
+	baseURI     string
```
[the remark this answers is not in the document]
MikeCJohnsonUk: Left over from cut'n'paste I suspect. Removed.

internal/clients/store/errors.go
```diff
@@ -55,3 +64,84 @@ type ErrStoreNotImplemented string
...
+func (eskff ErrStoreKeyFetchFailure) Error() string {
+	return fmt.Sprintf("CloudChamber: fetch txn failed reading key %q", string(eskff))
+}
+
+// ErrStoreKeyWriteFailure indicates the read transaction failed.
+//
+type ErrStoreKeyWriteFailure string
+
+func (eskwf ErrStoreKeyWriteFailure) Error() string {
+	return fmt.Sprintf("CloudChamber: fetch txn failed deleting key %q", string(eskwf))
```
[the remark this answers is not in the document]
MikeCJohnsonUk: Typos en mass. Fixed.

internal/clients/store/errors.go
```diff
@@ -55,3 +64,84 @@ type ErrStoreNotImplemented string
 func (esni ErrStoreNotImplemented) Error() string {
 	return fmt.Sprintf("CloudChamber: method %v not currently implemented", string(esni))
 }
+
+// ErrStoreKeyFetchFailure indicates the read transaction failed.
+//
+type ErrStoreKeyFetchFailure string
+
+func (eskff ErrStoreKeyFetchFailure) Error() string {
+	return fmt.Sprintf("CloudChamber: fetch txn failed reading key %q", string(eskff))
```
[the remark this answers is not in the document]
MikeCJohnsonUk: You know, elsewhere I made the opposite call for exactly those reasons? I'll expand it out.

internal/clients/store/store.go
```diff
@@ -959,3 +969,347 @@ func (store *Store) SetWatchWithPrefix(keyPrefix string) error {
...
+		response, err := concurrency.NewSTM(
+			store.Client,
+			actionReadRecords,
+			concurrency.WithIsolation(concurrency.ReadCommitted),
+			concurrency.WithPrefetch(keySet.Keys...),
+		)
+
+		if err != nil {
+			return err
+		} else if !response.Succeeded {
```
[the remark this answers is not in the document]
MikeCJohnsonUk: Done.

The things is I originally had it flatter, but then coalescing it along with the NewSTM call just aboce. I backed off on the call part as it just made everything look really _busy_, I just should have backed it all off. Sigh.

internal/clients/store/store.go
```diff
@@ -959,3 +969,347 @@ func (store *Store) SetWatchWithPrefix(keyPrefix string) error {
...
+		}
+	}
+
+	return nil
+}
+
+// ReadMultipleTxn is a method to fetch a set of arbitrary keys within a
+// single txn so they form a (self-)consistent set.
+//
+func (store *Store) ReadMultipleTxn(ctx context.Context, keySet RecordKeySet) (*RecordSet, error) {
```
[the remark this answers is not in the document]
MikeCJohnsonUk: It was originally even more so. I've already factored some of that out, The Read call is sufficiently different that it may not work cleanly there, but the Write and Delete calls are almost identical.

I need to properly get my head around lambdas and then I'll deal with this in a future change. I may even be able to do something with the Read call without too much wrangling.

internal/clients/store/storeapi.go
MikeCJohnsonUk: Fixed.

internal/clients/store/store.go
```diff
@@ -959,3 +969,347 @@ func (store *Store) SetWatchWithPrefix(keyPrefix string) error {
...
+// WriteMultipleTxn is a method to write/update a set of arbitrary keys within a
+// single txn so they form a (self-)consistent set.
+//
+func (store *Store) WriteMultipleTxn(ctx context.Context, recordSet *RecordUpdateSet) (int64, error) {
+
+	revision := RevisionInvalid
+
+	err := st.WithSpan(ctx, tracing.MethodName(1), func(ctx context.Context) (err error) {
+
+		if err := store.disconnected(ctx); err != nil {
```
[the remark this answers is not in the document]
MikeCJohnsonUk: Since technically it's in a different scope, there is no problem having another "err", it just occludes the one in the outer scope.

Just using "=" would work, you would then just be using the "err" in the outer scope until such time as it is overwritten when the span returns.

I'll admit to a preference with the new variable in the inner scope on the basis that it *is* a different error value with a different purpose, That it has the same name is incidental.

internal/clients/store/errors.go
```diff
@@ -55,3 +64,84 @@ type ErrStoreNotImplemented string
...
+	return fmt.Sprintf("CloudChamber: compare operator not valid for key %q", string(esbac))
+}
+
+// ErrStoreBadRecordCount indicates the record count for the operation was not valid.
+// This might mean that the store found more, or less, than the number of records expected.
+//
+type ErrStoreBadRecordCount string
+
+func (esbrc ErrStoreBadRecordCount) Error() string {
+	return fmt.Sprintf("CloudChamber: did not get the number of records expected %q", string(esbrc))
```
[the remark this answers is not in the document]
MikeCJohnsonUk: You know, I knew you were going to spot that one. At the time I created that error I had trouble creating a multi-parameter error. I'll fix that.

internal/clients/store/store.go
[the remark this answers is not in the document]
Jim3Things: Ok, I'm find with either choice.

internal/clients/store/storeapi.go
MikeCJohnsonUk: This results from a habit intended to improve readability in "if () {} else {}" statement situations where putting the "small" block at the top ensured it didn't get lost/swamped somewhere after the "big" block in the other arm. Only made sense really where one arm was *very* much larger than the other.

I've only rarely done this with switch() statements.

Given there is no fallthrough situations going on here, they are functionally equivalent.

I've only a minor preference in this case. If you've a strong preference I'll move it to the bottom.

Jim3Things: Not a strong preference, per se.  It is just different than the others, so it sort of hit me when reading it.

internal/clients/store/errors.go
```diff
@@ -55,3 +64,84 @@ type ErrStoreNotImplemented string
...
+func (eskdf ErrStoreKeyDeleteFailure) Error() string {
+	return fmt.Sprintf("CloudChamber: fetch txn failed deleting key %q", string(eskdf))
+}
+
+// ErrStoreWriteConditionFail indicates the update transaction failed due to a revision mismatch.
+//
+type ErrStoreWriteConditionFail string
+
+func (eswcf ErrStoreWriteConditionFail) Error() string {
+	return fmt.Sprintf("CloudChamber: condition fail/mismatch on update for key %q", string(eswcf))
```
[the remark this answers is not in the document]
MikeCJohnsonUk: I believe I've done that now. If you spot any that I missed....

internal/clients/store/errors.go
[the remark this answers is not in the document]
MikeCJohnsonUk: Fixed

internal/clients/store/store.go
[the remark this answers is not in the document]
MikeCJohnsonUk: Left as-is.

internal/clients/store/store.go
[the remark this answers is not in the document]
MikeCJohnsonUk: Left for a future change.

internal/clients/store/storeapi.go
MikeCJohnsonUk: Ok, I'll leave as is for now. If it becomes bothersome it's trivial to change after all.

internal/clients/store/errors.go
```diff
@@ -55,3 +51,102 @@ type ErrStoreNotImplemented string
...
+func (eskff ErrStoreKeyReadFailure) Error() string {
+	return fmt.Sprintf("CloudChamber: transaction failed reading key %q", string(eskff))
+}
+
+// ErrStoreKeyWriteFailure indicates the read transaction failed.
+//
+type ErrStoreKeyWriteFailure string
+
+func (eskwf ErrStoreKeyWriteFailure) Error() string {
+	return fmt.Sprintf("CloudChamber: transaction failed deleting key %q", string(eskwf))
```
Jim3Things, line 68: writing rather than deleting in the text?

internal/clients/store/store.go
```diff
@@ -959,3 +969,353 @@ func (store *Store) SetWatchWithPrefix(keyPrefix string) error {
...
+			concurrency.WithIsolation(concurrency.Serializable),
+			concurrency.WithPrefetch(*prefetchKeys...),
+		)
+
+		if err != nil {
+			return err
+		}
+
+		if !response.Succeeded {
+			err := ErrStoreKeyWriteFailure(recordSet.Label)
```
Jim3Things, line 1249: There's something about the handling of Label that I'm not following.  It may be that you should add a comment where it is declared in the type.

Label appears to be set to be the methodname of the function that created the record update set.  But I can only find it used here (so far).  And the text of the message suggests that it is a key.  I'm not sure how the method name turns into a key.

internal/clients/store/storeapi_test.go
```diff
@@ -0,0 +1,269 @@
...
+	assert.Nilf(t, err, "Failed to delete user %q - error: %v", userName, err)
+	assert.Lessf(t, revRead, revDelete, "Expected post-delete revision to be greater than read revision")
+
+	userReread, revReread, err := store.UserRead(context.Background(), userName)
+	assert.NotNilf(t, err, "Unexpected success reading user %q after deletion - error: %v", userName, err)
+	assert.Equalf(t, RevisionInvalid, revReread, "Unexpected difference in update revision vs read revision")
+	assert.Nilf(t, userReread, "Unexpected success in reading user %q after delete", userName)
+
+	revDeleteAgain, err := store.UserDelete(context.Background(), user, revRead)
+	assert.NotNilf(t, err, "Unecpected success trying to update with wrong revision for user %q - error: %v", userName, err)
```
Jim3Things, line 189: Unecpected => Unexpected

internal/clients/store/common_test.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	if cfg.Store.Test.PreCleanStore {
+		if err := cleanNamespace(testNamespace); err != nil {
+			log.Fatalf("failed to pre-clean the store as requested - namespace: %s err: %v", testNamespace, err)
+		}
+	}
+
+	setDefaultNamespaceSuffix(testNamespace)
+	return
```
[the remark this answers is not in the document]
MikeCJohnsonUk: Yup. Removed.

internal/clients/store/errors.go
MikeCJohnsonUk: Now how did that get there? Fixed.

internal/clients/store/storeapi_test.go
MikeCJohnsonUk: Fixed.

internal/clients/store/store.go
MikeCJohnsonUk: It's basically something I was using during debugging before I got the span stuff integrated. It's not actually required.

The Read and Delete routines have something similar.

The reason for it's use in that specific location is that there is the failure within the txn, so there isn't really a known applicable key (which key in the set would we choose?) so it is just using something common for the set, no better reason than that.

Partly the issue here is that I wanted to log _something_ and the txn itself did not fail, but the response out of the txn said it didn't succeed. I'm not entirely sure what this condition represents, so I needed something common to the set, and I happened to have this (debug) label, so...

The help text associated with this field is a little cryptic

// succeeded is set to true if the compare evaluated to true or false otherwise.

and 

// responses is a list of responses corresponding to the results from applying
// success if succeeded is true or failure if succeeded is false.

which seems to suggest it's applicable to the condition result. In our case, if the condition fails, we abort, so this error case may actually already be handled.


Anyway, I have some notion as to how to make this better involving factoring out much more of these primitives which will mean a change to the RecordUpdateSet struct, at which time I was planning on removing the label, unless I can think of a good reason to label the set beyond the calling functions name, which we have already. Perhaps a "reason" string?

Jim3Things: A reason string seems like a good approach, and how we got here makes sense to me.  I'm happy to have it logged as an issue and dealt with in the next refactoring.

MikeCJohnsonUk: So, I've just been doing some research on the behaviour of the Succeeded field, and what this field actually represents is the success or not, of the comparison used within the txn that forms the basis of the condition that we're using. Now in our case, this actually directly determines if the txn aborts, but theoretically at least, the flag can allow more complex failures to be described.

So, the outcome here.
1. I'll add the issue about the Succeeded flag and proper interpretation
2. I do quite like the notion of a "reason" though whether or not it will be useful long term I guess we'll have to see.

I'll regularize both in the next re-factoring.

Jim3Things: Ok

PR #84, opened 2020-07-15
Update users and tracing

Jim3Things:
- Restrict cookies to 'samesite=strict'

- Change all variants of the 'manage accounts' flag to be CanManageAccounts
- Add actual simulated time tick to the user trace entries

- Format stdout tracing to be closer to unit test tracing in format


internal/clients/store/storeapi_test.go
```diff
@@ -22,7 +22,7 @@ const (
 	alice                = "Alice"
 	bob                  = "Bob"
 	eve                  = "Eve"
-	alicePassword        = "AlicePassowrd"
+	alicePassword        = "AlicePassword"
```
MikeCJohnsonUk, line 25: Oh Jeez...

My typo I presume?

Jim3Things: Uhm, yeah...

PR #85, opened 2020-07-18
Replace frontend users dummy in-memory db with connection to the real store

MikeCJohnsonUk:
This update connects the frontend user db to the backend store to allow the persistence of records across service restarts etc. The dummy in-memory db was also removed and the tests updated to account for the persistence of records across test runs.

The store test namespace with pre-test cleanup was made available to all test cohorts, not just that for the store itself and the frontend tests were configured to use it to avoid interference between separate test runs.

Some issues with fuzzy layering were also fixed to ensure that each layer from users<->dbusers<->storeapi<->store were properly separated 

- user layer deals with http, http handlers, httperrors and userXxxx() routines, converts specific ErrUserXxxXxx to httperror for semantically significant cases.

- dbUsers deals with ErrUserXxxXxx errors, encodes and decodes pb.User structs to/from JSON buffers, and provides a CRUD style set of calls to the users layer, converts some specific ErrStoreXxxXxx to ErrUserXxxXxx errors for reportable situations

- storeapi provides a simplified CRUD style transactional api to the dbUsers layers, does name to/from key conversion, does NO encode/decode of values, returns ErrStoreXxxXxx errors.

- store provides the transactional primitives and interacts directly with the backend etcd store.

A few lint-er complaints were also fixed to reduce clutter in VsCode's "Problems" panel.

All tests run cleanly, even with multiple invocations. End-to-end use with the UI has not yet been verified.

### DIFF
--- a/internal/clients/store/errors.go
+++ b/internal/clients/store/errors.go
@@ -158,3 +158,11 @@ type ErrStoreAlreadyExists string
 func (esae ErrStoreAlreadyExists) Error() string {
 	return fmt.Sprintf("CloudChamber: condition failure (already exists) on create for key %q", string(esae))
 }
+
+// ErrStoreInvalidConfiguration indicates the key, value pair being created already exists
+//
+type ErrStoreInvalidConfiguration string
+
+func (esic ErrStoreInvalidConfiguration) Error() string {
+	return fmt.Sprintf("CloudChamber: invalid store configuration - %v", string(esic))
+}
--- a/internal/clients/store/errors.go
+++ b/internal/clients/store/errors.go
@@ -60,7 +60,7 @@ func (eskff ErrStoreKeyReadFailure) Error() string {
 	return fmt.Sprintf("CloudChamber: transaction failed reading key %q", string(eskff))
 }
 
-// ErrStoreKeyWriteFailure indicates the read transaction failed.
+// ErrStoreKeyWriteFailure indicates the write transaction failed as a result of a pre-condition failure.
 //
 type ErrStoreKeyWriteFailure string
 
@@ -97,7 +97,7 @@ func (esbrk ErrStoreBadRecordKey) Error() string {
 type ErrStoreBadRecordContent string
 
 func (esbrk ErrStoreBadRecordContent) Error() string {
-	return fmt.Sprintf("CloudChamber: discovered found record where content does not match key %q", string(esbrk))
+	return fmt.Sprintf("CloudChamber: discovered record where content does not match key %q", string(esbrk))
 }
 
 // ErrStoreBadRecordCount indicates the record count for the operation was not valid.
@@ -149,4 +149,12 @@ type ErrStoreConditionFail struct {
 
 func (esucf ErrStoreConditionFail) Error() string {
 	return fmt.Sprintf("CloudChamber: condition failure on update for key %q - requested: %v condition: %v actual: %v", esucf.key, esucf.requested, esucf.condition, esucf.actual)
+}
+
+// ErrStoreRecordExists indicates the key, value pair being created already exists
+//
+type ErrStoreRecordExists string
+
+func (esre ErrStoreRecordExists) Error() string {
+	return fmt.Sprintf("CloudChamber: condition failure (already exists) on create for key %q", string(esre))
 }
--- a/internal/clients/store/errors.go
+++ b/internal/clients/store/errors.go
@@ -151,10 +151,10 @@ func (esucf ErrStoreConditionFail) Error() string {
 	return fmt.Sprintf("CloudChamber: condition failure on update for key %q - requested: %v condition: %v actual: %v", esucf.key, esucf.requested, esucf.condition, esucf.actual)
 }
 
-// ErrStoreRecordExists indicates the key, value pair being created already exists
+// ErrStoreAlreadyExists indicates the key, value pair being created already exists
 //
-type ErrStoreRecordExists string
+type ErrStoreAlreadyExists string
 
-func (esre ErrStoreRecordExists) Error() string {
-	return fmt.Sprintf("CloudChamber: condition failure (already exists) on create for key %q", string(esre))
+func (esae ErrStoreAlreadyExists) Error() string {
+	return fmt.Sprintf("CloudChamber: condition failure (already exists) on create for key %q", string(esae))
 }
--- a/internal/clients/store/store.go
+++ b/internal/clients/store/store.go
@@ -1089,7 +1089,7 @@ func checkConditions(stm concurrency.STM, recordSet *RecordUpdateSet) error {
 	for k, ru := range recordSet.Records {
 		if ru.Condition == ConditionCreate {
 			if stm.Get(k) != "" {
-				return ErrStoreConditionFail{k, RevisionInvalid, ru.Condition, stm.Rev(k)}
+				return ErrStoreRecordExists(k)
 			}
 		} else if ru.Condition != ConditionUnconditional {
 			rev := stm.Rev(k)
@@ -1123,6 +1123,101 @@ func checkConditions(stm concurrency.STM, recordSet *RecordUpdateSet) error {
 			case ConditionRevisionGreater:
 				if rev <= ru.Record.Revision {
 					return ErrStoreConditionFail{k, ru.Record.Revision, ru.Condition, rev}
+				}
+			}
+		}
+	}
+
+	return nil
+}
+
+func getPrefetchKeys(req *Request) (*[]string, error) {
+
+	prefetchKeys := make([]string, 0, len(req.Records))
+
+	// Build an array of keys to supply as the arg to prefetch
+	// on the WithPrefetch() option below
+	//
+	for k, r := range req.Records {
+		c := req.Conditions[k]
+		switch c {
+		case ConditionUnconditional:
+			// No need to prefetch if not comparing anything
+			break
+
+		case ConditionRevisionNotEqual:
+			fallthrough
+
+		case ConditionRevisionLess:
+			fallthrough
+
+		case ConditionRevisionLessOrEqual:
+			fallthrough
+
+		case ConditionRevisionEqual:
+			fallthrough
+
+		case ConditionRevisionEqualOrGreater:
+			fallthrough
+
+		case ConditionRevisionGreater:
+			if r.Revision == RevisionInvalid {
+				return nil, ErrStoreBadArgRevision{k, RevisionInvalid, r.Revision}
+			}
+
+			fallthrough
+
+		case ConditionCreate:
+			prefetchKeys = append(prefetchKeys, k)
+
+		default:
+			return nil, ErrStoreBadArgCondition{k, Condition(c)}
+		}
+	}
+
+	return &prefetchKeys, nil
+}
+
+func chkConditions(stm concurrency.STM, req *Request) error {
+
+	for k, r := range req.Records {
+		c := req.Conditions[k]
+		if c == ConditionCreate {
+			if stm.Get(k) != "" {
+				return ErrStoreRecordExists(k)
+			}
+		} else if c != ConditionUnconditional {
+			rev := stm.Rev(k)
+
+			switch c {
+			case ConditionRevisionLess:
+				if rev >= r.Revision {
+					return ErrStoreConditionFail{k, r.Revision, c, rev}
+				}
+
+			case ConditionRevisionLessOrEqual:
+				if rev > r.Revision {
+					return ErrStoreConditionFail{k, r.Revision, c, rev}
+				}
+
+			case ConditionRevisionEqual:
+				if rev != r.Revision {
+					return ErrStoreConditionFail{k, r.Revision, c, rev}
+				}
+
+			case ConditionRevisionNotEqual:
+				if rev == r.Revision {
+					return ErrStoreConditionFail{k, r.Revision, c, rev}
+				}
+
+			case ConditionRevisionEqualOrGreater:
+				if rev < r.Revision {
+					return ErrStoreConditionFail{k, r.Revision, c, rev}
+				}
+
+			case ConditionRevisionGreater:
+				if rev <= r.Revision {
+					return ErrStoreConditionFail{k, r.Revision, c, rev}
 				}
 			}
 		}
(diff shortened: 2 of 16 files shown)


internal/clients/store/store.go
```diff
@@ -1131,6 +1218,158 @@ func checkConditions(stm concurrency.STM, recordSet *RecordUpdateSet) error {
...
+		if c == ConditionCreate {
+			if stm.Get(k) != "" {
+				return ErrStoreAlreadyExists(k)
+			}
+		} else if c != ConditionUnconditional {
+			rev := stm.Rev(k)
+
+			switch c {
+			case ConditionRevisionLess:
+				if rev >= r.Revision {
```
Jim3Things, line 1281: I suggested this earlier, but want to add that this would be easier to read if the comparisons were reversed.  That way the operator would match the condition name.  Functionally the same, just involves less internal reversals...

internal/clients/store/store.go
```diff
@@ -1131,6 +1218,158 @@ func checkConditions(stm concurrency.STM, recordSet *RecordUpdateSet) error {
...
+// which have a common prefix. The method will return all matching key/value pairs so
+// care should be taken with key naming to avoid attempting to fetch a large number
+// of key/value pairs.
+//
+// It is not an error to attmept to retrieve an empty set. For example, when querying
+// for the presence of a set of values, this method can be used which would successfully
+// return an empty set of key/value pairs if there are no matches for the supplied key
+// prefix.
+//
+func (store *Store) ListWithPrefix(ctx context.Context, keyPrefix string) (response *Response, err error) {
```
Jim3Things, line 1326: Personally, I'd be happier if all the returns used the two variable form.  The fact that you've named these, I'm guessing, gives you a path to avoid it, but it seems confusing to read.

internal/clients/store/store.go
```diff
@@ -1131,6 +1218,158 @@ func checkConditions(stm concurrency.STM, recordSet *RecordUpdateSet) error {
...
+			}
+		}
+
+		resp.Revision = getResponse.Header.GetRevision()
+
+		response = resp
+
+		st.Infof(ctx, -1, "Processed %v items", len(resp.Records))
+
+		return nil
```
Jim3Things, line 1367: This would be shorter and clearer (at least to me) if it were line 1365, followed by "return resp, nil"

I'm guessing this is a style comment that repeats...

internal/clients/store/storeapi.go
```diff
@@ -184,140 +264,205 @@ func (store *Store) UserDelete(ctx context.Context, u *pb.User, revision int64)
...
+
+		return nil
+	})
+
+	return revision, err
 }
 
-// UserRecordSet is a collection of UserRecord structs used to describe
-// a set of users all with a common store revision.
+// DeleteNew is a function to create a single key, value record pair
```
Jim3Things, line 369: create => delete?

internal/services/frontend/DBUsers.go
```diff
@@ -12,146 +12,243 @@
...
+		Name:              cfg.WebServer.SystemAccount,
+		PasswordHash:      passwordHash,
+		Enabled:           true,
+		CanManageAccounts: true,
+		NeverDelete:       true})
+
+	// If the SystemAccount already exists, eat the failure
+	//
+	if err == ErrUserAlreadyExists(cfg.WebServer.SystemAccount) {
+		return nil
```
Jim3Things, line 64: There are actually two cases here - one where the admin name exists and the admin password matches that in the config file, and the other where the admin name exists but the password does not match what is in the config file.  In both cases I think we should continue to come up, but in the second we should put an informational or warning into the trace so that the user can understand why it came up but they couldn't log in, and what to do about it.

Since this method doesn't trace, I think this means a second error code that the caller interprets and logs.

internal/services/frontend/DBUsers.go
```diff
@@ -12,146 +12,243 @@
...
-    m.Mutex.Lock()
-    defer m.Mutex.Unlock()
+	v, err := store.Encode(u)
+
+	if err != nil {
+		return InvalidRev, err
+	}
 
-    key := strings.ToLower(name)
+	rev, err := m.Store.CreateNewValue(context.Background(), store.KeyRootUsers, u.Name, v)
```
Jim3Things, line 81: Does something still lowercase the name?  I'm sure it does, but I'm a little lost as to where.

internal/services/frontend/users_test.go
```diff
@@ -563,7 +563,8 @@ func TestUsersRead(t *testing.T) {
 	assert.Nilf(t, err, "Failed to convert body to valid json.  err: %v", err)
 
 	assert.Equal(t, "application/json", strings.ToLower(response.Header.Get("Content-Type")))
-	assert.Equal(t, "1", response.Header.Get("ETag"))
+	// TODO MCJ - false expectation for E-Tag assert.Equal(t, "1", response.Header.Get("ETag"))
+	assert.Less(t, "1", response.Header.Get("ETag"))
```
Jim3Things, line 567: I don't think this form of the test works reliably.  It will fail, I believe, if either rev count is >9.  e.g. "10" is probably less than "2".  I think you need to convert the ETag to a number and do this test using numeric values.

... and assume comment applies to all instances...

internal/clients/store/store.go
```diff
@@ -238,10 +242,92 @@ func Initialize(cfg *config.GlobalConfig) {
...
 			storeRoot.DefaultTraceFlags,
 			storeRoot.DefaultNamespaceSuffix)
+
+		// See if any part of the test namespace requires initialization and optionally cleanining.
+		//
+		// NOTE: This only affects the test namespace, not the production namespace. Ideally it
+		//       would not be here but the store is used by other services as part of their test
+		//       operation and so this feature needs to be in common code.
+		//
+		PrepareTestNamespace(ctx, cfg)
```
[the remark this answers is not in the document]
MikeCJohnsonUk: Ah. Clearly I need to better explain this.

The use-for-test config flag is basically what I just added. The store test code explicitly did this for itself in the earlier versions because it was test code. By moving this "feature" into the main body there is no longer the implicit determination of the use-case by virtue of it being in the test code so I added the "UseTestNamespace" config flag.
What this does is change the root namespace prefix on all the keys we write to the store. This is done invisibly to the client of the store.

For production, all our (persisted) keys have a prefix of "/CloudChamber/V0.1"
If the config flag UseTestNamespace is set to true (defaults to false obviously ;-), then the prefix becomes "/CloudChamber/V0.1/Test/Xxxx" where the "Xxxx" portion is either "Standard" if "UseUniqueInstance" is false, or a timestamp if "UseUniqueInstance" is true.

These prefixes give a number of useful behaviours. By having a project specific prefix we should be able to share any etcd insatance, assuming of course that no one else uses the same prefix. By having a well defined test prefix, it becomes easy to monitor, control and also of course test data does not interfere with any current production data.

Also, by using "/CloudChamber/V0.1/Test" it becomes very easy to purge the store any any and all test related records. And that is what happens if you set the "PreCleanStore" config flag (actually it only auto-cleans the "Standard" leg of the namespace but you get the idea). 

Additionally, it is easy to delete all the keys with a common prefix manually via the etcdctl utility without having to disturb the production data at all.

I'll place, and expand upon, this text into a readme in the store directory.

internal/clients/store/store.go
```diff
@@ -238,10 +242,92 @@ func Initialize(cfg *config.GlobalConfig) {
...
+// clean out the store of data left from previous runs and set a test specific namespace for all
+// subsequent store operations.
+//
+// NOTE: It is expected that this is call soon after the store is initiaitlized and before any
+//       data related operations have taken place.
+//
+func PrepareTestNamespace(ctx context.Context, cfg *config.GlobalConfig) {
+
+	if !cfg.Store.Test.UseTestNamespace {
+		return
```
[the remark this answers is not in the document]
MikeCJohnsonUk: We already do and allow this behaviour. Whether or not we want to retain the ability for test systems and production systems to share the same store is more of a deployment choice. We have the ability. We do not need to use it.

internal/clients/store/store.go
```diff
@@ -238,10 +242,92 @@ func Initialize(cfg *config.GlobalConfig) {
...
+	// For test purposes, need to set an alternate namespace rather than
+	// rely on the standard. From the configuration, we can either use the
+	// standard, fixed, well-known prefix, or we can use a per-instance
+	// unique prefix derived from the current time
+	//
+	testNamespace := testNamespaceSuffix
+
+	if cfg.Store.Test.UseUniqueInstance {
+		testNamespace += fmt.Sprintf("/%s", time.Now().Format(time.RFC3339Nano))
+	} else {
```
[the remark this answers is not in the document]
MikeCJohnsonUk: A "namespace" is really just an automatic prefixing of the keys being use. Is there any overhead? Some, as a result of having to add/remove the prefix on all request/response interactions, but other than that, I don't think so. 

The current code already allows the config to enable automatically cleaning out the "/CloudChamber/V0.1/Test/Standard" namespace.

Optionally we can use the etcdctl utility to do the same. The ability is already covered in the "Operational Environment" doc although that does not identify the specific namespace it might need to apply to. I was already planning a couple of scripts for this in a separate checkin.

internal/clients/store/store.go
MikeCJohnsonUk: Not a bad idea. I admit did have to stop and think as I wrote those lines so there is justification.

I'll do this.

internal/clients/store/store.go
MikeCJohnsonUk: This is a result of being inside a trace span, so there is only the error that can be returned directly. Any other value(s) are returned outside the scope of the span where the parent (enclosing) function returns them by name.

internal/clients/store/storeapi.go
MikeCJohnsonUk: Oops. 

Fixed.

internal/services/frontend/DBUsers.go
MikeCJohnsonUk: Yes.

I moved all the use of the supplied name as part of a key into the storeapi layer in a set of routines with names along the lines of getKeyFromXxxXxx(), the most used of which is getKeyFromKeyRootAndName(). This takes the supplied key root which identifies whether this is for a user, a workload or inventory item etc and uses this to generate a prefix to the normalized name. The preparation of a name for use as part of a key is done by GetNormalizedName(). Currently all the prep that is required is lowercasing the supplied string.

This can be a little confusing, but came about because of trying to test for things like ErrXxxXxx() using a name, and making sure that we always used names of the same case for keys etc. Especially when it comes to what we report from List() where we report names back to the uppers layers using the names from within the records, i.e. there stored values.

Basically, the "application" layers (dbusers, dbinventory etc) all talk in terms of names using whatever makes sense to them. The derivation of a store key from the supplied name happens at the storeapi layer.

Fortunately we never have to reverse this key derivation procedure: it's only ever one-way.

internal/services/frontend/users_test.go
MikeCJohnsonUk: Ah. Yes, I hadn't actually considered beyond getting the assert() to not trigger. We currently get away with it but I'll move the comparison into the numeric realm.

I'll fix this.

internal/services/frontend/DBUsers.go
MikeCJohnsonUk: Hhmm. We could actually just do the trace at this level. Right now the caller has no real idea as to what initialization requires and to add a specific error code which all callers (currently two of them) would then have to interpret and react to could easily be done right where the account creation happens. 

I'll add the tracing and checking at this location unless you'd prefer it elsewhere.

internal/clients/store/store.go
MikeCJohnsonUk: Not sure I understand.

Is this part of the missing-context-in-a-review issue we were talking about with regards to the returnable values within a trace span?

internal/clients/store/store.go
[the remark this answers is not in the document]
Jim3Things: Ok, that would be great.  Thanks.

internal/clients/store/store.go
Jim3Things: Ah, ok makes sense.  Reading issue on my part.  I'll claim the diff tool, but... :)

internal/services/frontend/DBUsers.go
Jim3Things: No, here is fine.

internal/clients/store/store.go
Jim3Things: Yup.  You can ignore this.

internal/clients/store/errors.go
```diff
@@ -150,3 +150,19 @@ type ErrStoreConditionFail struct {
...
 	return fmt.Sprintf("CloudChamber: condition failure on update for key %q - requested: %v condition: %v actual: %v", esucf.key, esucf.requested, esucf.condition, esucf.actual)
 }
+
+// ErrStoreAlreadyExists indicates the key, value pair being created already exists
+//
+type ErrStoreAlreadyExists string
+
+func (esae ErrStoreAlreadyExists) Error() string {
+	return fmt.Sprintf("CloudChamber: condition failure (already exists) on create for key %q", string(esae))
+}
```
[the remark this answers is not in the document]
MikeCJohnsonUk: Within a txn, we can certainly access the current persisted value and consequently we can compare and abort to avoid an unnecessary write. That would even be easy to do within the confines of the txn. There would need to be a means of describing the requirement, but it would be something along the lines of what I was already planning for the conditions for things like Delete()  to avoid the 3-part RMW. Basically we add a condition of "only update if value different" and we should even be able to return the more recent revision of the current version. How does that sound? I mean yes, it's an optimization, but then so is the "RMW into conditional update" I'm already planning on.

internal/clients/store/store.go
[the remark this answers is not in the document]
MikeCJohnsonUk: We probably need some text somewhere to explain all the config options somewhere as to what they do, acceptable values etc.

internal/clients/store/store.go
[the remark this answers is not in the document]
MikeCJohnsonUk: We probably need some text somewhere to explain all the config options somewhere as to what they do, acceptable values etc.

internal/services/frontend/DBUsers.go
MikeCJohnsonUk: Fixed.

internal/services/frontend/users_test.go
MikeCJohnsonUk: Fixed.

internal/clients/store/store.go
MikeCJohnsonUk: Hhmm, on second thoughts....

The naming here is obscuring things.

rev represents the current revision of the record in the store
r.Revision is the callers supplied revision, i.e. the revision they think they are working with.

So what this is really saying is

if the callers version (r.Revision) is older than the store version, then fail.

It's not a straight reversal. Or at least it looks like it is, but it's not. We're really looking for the failure condition.

Consider the "same" case on line 1290. We are going to succeed if rev == r.Revision, or in other words fail if they are not equal.

I'm not saying that this couldn't be made more obvious, it's just not as obvious as it looks.

So the first one should be really seen as

case ConditionRevisionLess:
  fail if rev !< r.Revision

The ConditionXxxx is about success, the ">=" is about selecting on NOT success. Since we don't have "!<", the not less-than is greater or equal.

Confusing right?


Rather than describing the ConditionXxxXxx as less-than etc, would it help to change to earlier, later, or newer, older and so on?

And perhaps replace rev with storeRev and r.Revision with callerRev

Then we might have

case ConditionRequireEarlierRevision:
if storeRev >= callerRev return ErrXxxx()

case ConditionRequireSameRevision:
if storeRev != callerRev return ErrXxx()

etc.


To be honest, so far I've only used (and hence tested) Unconditional and Equal. It's not clear to me the other options are even required. I only really added them for completeness. The simplest thing here might be to just remove them.

I'm open the suggestions here.

internal/clients/store/errors.go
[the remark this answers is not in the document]
MikeCJohnsonUk: Future checkin.

internal/services/frontend/DBUsers.go
MikeCJohnsonUk: Fixed.

internal/services/frontend/users_test.go
MikeCJohnsonUk: Fixed

internal/clients/store/store.go
Jim3Things: I'll leave this to your analysis.  I'm open to different naming or restricting the set of checks that we support.  I just found reading it harder than expected due to mental gyrations I was doing, and am looking for a way to avoid them.

PR #92, opened 2020-07-23
Users changes for update, set password

Jim3Things:
- Also a fix to not extend the rights for a user beyond what the caller has.

### DIFF
--- a/internal/services/frontend/DBUsers.go
+++ b/internal/services/frontend/DBUsers.go
@@ -13,23 +13,20 @@ package frontend
 
 import (
 	"context"
-	"sync"
+
+	"golang.org/x/crypto/bcrypt"
 
 	"github.com/Jim3Things/CloudChamber/internal/clients/store"
 	"github.com/Jim3Things/CloudChamber/internal/config"
 	"github.com/Jim3Things/CloudChamber/internal/tracing"
 	st "github.com/Jim3Things/CloudChamber/internal/tracing/server"
 	pb "github.com/Jim3Things/CloudChamber/pkg/protos/admin"
-	"golang.org/x/crypto/bcrypt"
 )
 
 // DBUsers is a container used to established synchronized access to
 // the in-memory set of user records.
 //
 type DBUsers struct {
-	Mutex sync.Mutex
-	Users map[string]*pb.UserInternal
-
 	Store *store.Store
 }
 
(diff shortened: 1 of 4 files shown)


internal/services/frontend/DBUsers.go
```diff
@@ -149,47 +144,94 @@ func (m *DBUsers) Read(name string) (*pb.User, int64, error) {
...
 
 	if err != nil {
-		return InvalidRev, err
+		return nil, InvalidRev, err
 	}
 
 	if rev != match {
-		return InvalidRev, ErrUserStaleVersion(u.Name)
+		return nil, InvalidRev, ErrUserStaleVersion(name)
 	}
```
MikeCJohnsonUk, line 161: Technically, we don't need this, as the same check will effectively happen as a pre-condition in the m.Store.UpdateNew() call. It does represent an optimization though so not harmful.

I am looking at providing the ability to feed an action routine into calls for UpdateXxx() and DeleteXxx() in particular which will allow the ReadNewValue() call to go away and will allow the call here for example, to do the old password matching within the bound of the underlying Update txn, i.e. make this a conditional update based upon the caller-supplied condition.

Until then however, this is the best we can do.

internal/services/frontend/users.go
```diff
@@ -56,13 +56,14 @@ func usersAddRoutes(routeBase *mux.Router) {
...
 
 	// Routes for individual user operations
-	routeUsers.HandleFunc(routeString, handlerUsersCreate).Methods("POST")
-	routeUsers.HandleFunc(routeString, handlerUsersRead).Methods("GET")
-	routeUsers.HandleFunc(routeString, handlerUsersUpdate).Methods("PUT")
-	routeUsers.HandleFunc(routeString, handlerUsersDelete).Methods("DELETE")
+	routeUsers.HandleFunc(routeString, handlerUserCreate).Methods("POST")
+	routeUsers.HandleFunc(routeString, handlerUserRead).Methods("GET")
+	routeUsers.HandleFunc(routeString, handlerUserUpdate).Methods("PUT")
+	routeUsers.HandleFunc(routeString, handlerUserDelete).Methods("DELETE")
```
MikeCJohnsonUk, line 66: Ah, that reads better. Good call...

internal/services/frontend/DBUsers.go
Jim3Things: Cool.  Look forward to that change.

PR #93, opened 2020-07-23
Cleanup some store tracing and assorted VsCode lint-er complaints

MikeCJohnsonUk:
Removed some redundant tracing in storeapi
Added some intent type trace entries (until "explainer" tracing arrives)
Sorted out a lot of complaints from lint in VsCode, mainly comment starting symbol matching public symbols and the insistence that Json should be all upper-case.

No functional changes. Tested clean.

### DIFF
--- a/internal/services/frontend/stepper_test.go
+++ b/internal/services/frontend/stepper_test.go
@@ -1,371 +1,371 @@
 package frontend
 
 import (
-    "fmt"
-    "net/http"
-    "net/http/httptest"
-    "testing"
-    "time"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
 
-    "github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/assert"
 
-    ts "github.com/Jim3Things/CloudChamber/internal/clients/timestamp"
-    "github.com/Jim3Things/CloudChamber/internal/tracing/exporters/unit_test"
-    pb "github.com/Jim3Things/CloudChamber/pkg/protos/Stepper"
-    "github.com/Jim3Things/CloudChamber/pkg/protos/common"
+	ts "github.com/Jim3Things/CloudChamber/internal/clients/timestamp"
+	"github.com/Jim3Things/CloudChamber/internal/tracing/exporters/unit_test"
+	pb "github.com/Jim3Things/CloudChamber/pkg/protos/Stepper"
+	"github.com/Jim3Things/CloudChamber/pkg/protos/common"
 )
 
 const (
-	stepperURI    = "/api/stepper"
+	stepperURI = "/api/stepper"
 )
 
 func testStepperPath() string { return baseURI + stepperURI }
 
 func testStepperReset(t *testing.T) {
-    err := ts.Reset()
-    assert.Nilf(t, err, "Unexpected error when resetting the stepper service)")
+	err := ts.Reset()
+	assert.Nilf(t, err, "Unexpected error when resetting the stepper service)")
 }
 
 func testStepperSetManual(t *testing.T, match int64, cookies []*http.Cookie) []*http.Cookie {
-    request := httptest.NewRequest("PUT", fmt.Sprintf("%s%s", testStepperPath(), "?mode=manual"), nil)
-    request.Header.Set("If-Match", fmt.Sprintf("%v", match))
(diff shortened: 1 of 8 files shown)


internal/services/frontend/stepper_test.go
```diff
@@ -1,371 +1,371 @@
...
-    assert.Equal(t, int64(0), res.Now.Ticks, "Unexpected current time")
-    assert.Equal(t, int64(0), res.WaiterCount, "Unexpected active waiter count")
-    assert.Equal(t, stat.Epoch + 1, res.Epoch, "Unexpected epoch value")
+	res, cookies := testStepperGetStatus(t, cookies)
+	assert.Equal(t, pb.StepperPolicy_Manual, res.Policy, "Unexpected policy")
+	assert.Equal(t, int64(0), res.MeasuredDelay.Seconds, "Unexpected delay")
+	assert.Equal(t, int32(0), res.MeasuredDelay.Nanos, "Unexpected delay")
+	assert.Equal(t, int64(0), res.Now.Ticks, "Unexpected current time")
+	assert.Equal(t, int64(0), res.WaiterCount, "Unexpected active waiter count")
+	assert.Equal(t, stat.Epoch+1, res.Epoch, "Unexpected epoch value")
```
Jim3Things, line 124: I'm not happy with this change - is this a settable choice in the linter (I"m assuming that is where it came from)?

internal/services/frontend/stepper_test.go
```diff
@@ -1,371 +1,371 @@
...
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+
+	ts "github.com/Jim3Things/CloudChamber/internal/clients/timestamp"
+	"github.com/Jim3Things/CloudChamber/internal/tracing/exporters/unit_test"
+	pb "github.com/Jim3Things/CloudChamber/pkg/protos/Stepper"
+	"github.com/Jim3Things/CloudChamber/pkg/protos/common"
```
Jim3Things, line 15: Have you found out what is going on with the whitespace?  It looks like there's something about tabs vs spaces, if I had to guess.  And I wonder if the normal go pattern of 2 spaces / indent vs our 4 spaces / indent is messing with us in some way?

internal/services/frontend/stepper_test.go
MikeCJohnsonUk: Err, I didn't explicitly change this.

This is probably the VsCode "prettifier", at least the leading whitespace was.

Is it the spacing around the "+" that you're not happy about in particular? I'm not a fan of the change and given the choice I would have left (and preferred) it as " + ".

internal/services/frontend/stepper_test.go
MikeCJohnsonUk: Nope, not yet.

It's looking like it might be time to finally track it down. I know it likes to play with leading 4-spaces into tabs and vice-versa. It also like removing (some) spaces in expressions. Not sure why some and not others though: it seems somewhat arbitrary.

internal/services/frontend/stepper_test.go
Jim3Things: Yeah, it was the spaces around the '+'.  Looks like we agree.  Let's see if we can find out what is causing this...

internal/services/frontend/stepper_test.go
MikeCJohnsonUk: Well, I've found, and fixed, what is probably the culprit. I did have use spaces and tabs set to 4 spaces, but there is this other setting, "DetectIndentation" which *was* enabled, and helpfully overrides the tabs-to-spaces and size values.

I hate "smart" software...

The only thing I've found so far on the spaces around operators is specific to typescript. Still looking...

MikeCJohnsonUk: I think I may have a handle on this, or at least in part.

The go extension seems it really, *REALLY* wants to auto detect indentation, and even tough I'd disable it in the global user settings, VsCode turned it back on for the project level settings. So it seems I need to turn it off their too since it doesn't seem to want to accept the word of the global user setting. Sigh.

The only thing that might affect the spaces I can see right now is to disable format-on-save. That seems to a a "Leave it TF alone" option. We'll see.

MikeCJohnsonUk:
Superseded by changes from branch mj-cleanup-2

PR #95, opened 2020-07-25
InventorySessionTestsCompleted25Jul20

manpreet85:
One negative test was added to check that the Inventory functions runs in a valid & established session only.


internal/services/frontend/inventory_test.go
```diff
@@ -191,4 +233,16 @@ func TestInventoryBlade2Read(t *testing.T) {
...
+
+// The purpose of this test is to check that the Inventory function get executed in a valid & established http session only
+func TestInventoryNoSession(t *testing.T) {
+	unit_test.SetTesting(t)
+	defer unit_test.SetTesting(nil)
+
+	request := httptest.NewRequest("GET", "/api/racks/", nil)
+	response := doHTTP(request, nil)
+
+	assert.Equal(t, http.StatusInternalServerError, response.StatusCode, "Handler returned Internal server error: %v", response.StatusCode)
```
Jim3Things, line 247: This is good, thanks.

There should be four in total - one for each of the handlers.  I'm ok with filing an issue for the other three and fixing them in a later checkin.  I'll do that just now.

I'm also going to file an issue on myself about the failure status code - I don't like that we're getting a 500.

PR #113, opened 2020-08-08
Streamline the memo state

Jim3Things:
- Removed the stale memo state entries from the ExternalRack message
- Removed the redundant rack name from the ExternalRackSummary message (the name is held as the key in the map).  Retained the ExternalRackSummary struct for future expansion.
- Moved the Zone memo fields out of the proto and into the DBInventory struct definition.  This reflects that the memo data is purely internal and in-memory, so there is no need for it to be in the proto.

### DIFF
--- a/internal/services/frontend/DBInventory.go
+++ b/internal/services/frontend/DBInventory.go
@@ -26,6 +26,8 @@ import (
 type DBInventory struct {
 	Mutex sync.Mutex
 	Zone *pb.ExternalZone
+	MaxBladeCount int64
+	MaxCapacity *common.BladeCapacity
 }
 
 var dbInventory *DBInventory
@@ -42,9 +44,10 @@ func InitDBInventory() error {
 		dbInventory = &DBInventory{
 			Mutex: sync.Mutex{},
 			Zone: &pb.ExternalZone{
-				Racks:         make(map[string]*pb.ExternalRack),
-				MaxCapacity: &common.BladeCapacity{},
+				Racks:       make(map[string]*pb.ExternalRack),
 			},
+			MaxBladeCount: 0,
+			MaxCapacity: &common.BladeCapacity{},
 		}
 
 		dbInventory.Zone.Racks["rack1"] = &pb.ExternalRack{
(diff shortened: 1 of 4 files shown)


internal/services/frontend/DBInventory.go
```diff
@@ -26,6 +26,8 @@ import (
 type DBInventory struct {
 	Mutex sync.Mutex
 	Zone *pb.ExternalZone
+	MaxBladeCount int64
+	MaxCapacity *common.BladeCapacity
```
MikeCJohnsonUk, line 30: So for the entire inventory, not just for a zone, assuming their might be more than one zone at some point?

internal/services/frontend/DBInventory.go
```diff
@@ -181,11 +184,10 @@ func (m *DBInventory) GetBlade(rackID string, bladeID int64) (*common.BladeCapac
 // buildSummary constructs the memo-ed summary data for the zone.  This should
 // be called whenever the configured inventory changes.
 func (m *DBInventory) buildSummary() {
-	m.Zone.MaxBladeCount = 0
+	m.MaxBladeCount = 0
```
MikeCJohnsonUk, line 187: Strictly speaking, not yet necessary, but what do we envisage for the locking strategy here? call invocation site or target?

pkg/protos/inventory/external.proto
```diff
@@ -38,18 +38,11 @@ message external {
         // specify the blades in the rack.  Each blade is defined by an integer index within that rack, which is used
         // here as the key.
         map<int64, common.blade_capacity> blades = 3;
-
-        // max_blade_capacity is a cached summary of the maximum values for
-        // each of the blade capacity values.  This can be used to size any
-        // collective values, such as having the UI frame the zone visually
```
MikeCJohnsonUk, line 44: Any chance we could move this comment either to the field defn in dbInventory, or perhaps on the GetMemoDate() func?

Unless it appears somewhere else that I failed to spot/realize.

internal/services/frontend/DBInventory.go
Jim3Things: Correct

PR #116, opened 2020-08-11
Bugfix - set correct status code for no active session

Jim3Things:


internal/services/frontend/inventory_test.go
```diff
@@ -254,23 +260,25 @@ func TestInventoryNoSession(t *testing.T) {
 	request := httptest.NewRequest("GET", "/api/racks/", nil)
 	response := doHTTP(request, nil)
 
-	assert.Equal(t, http.StatusInternalServerError, response.StatusCode, "Handler returned Internal server error: %v", response.StatusCode)
+	assert.Equal(t, http.StatusForbidden, response.StatusCode, "Handler returned Internal server error: %v", response.StatusCode)
```
MikeCJohnsonUk, line 263: Text no longer matches expected status response?

Same comment applies to further updates further on down.

internal/services/frontend/session_manager.go
```diff
@@ -206,11 +205,13 @@ func doSessionHeader(
 
 // ensureEstablishedSession verifies that the session is not new, and triggers
 // an error if it is.
-func ensureEstablishedSession(ctx context.Context, session *sessions.Session) error {
+func ensureEstablishedSession(session *sessions.Session) error {
 	if session.IsNew {
-		return st.Errorf(
-			ctx, -1,
-			"Unexpected new session, value=%s", dumpSessionState(session))
+		return NewErrNoSessionActive()
```
MikeCJohnsonUk, line 210: Is this the appropriate error seeing that the surrounding code implies there is a session, it's just that it's new vs previously established.

internal/services/frontend/inventory_test.go
Jim3Things: Fixed.

internal/services/frontend/session_manager.go
Jim3Things: Yeah, that's an artifact of how gorilla handles looking up a session that doesn't exist - it creates a new one.  So by this point you can only tell if there was no session when called by looking at the IsNew flag

MikeCJohnsonUk: Got it.

PR #117, opened 2020-08-12
Minor frontend additions

Jim3Things:
- Add tracing for file retrieval requests
- Add a /ping route that is used to refresh the CC-session inactivity timer

### DIFF
--- a/internal/services/frontend/frontend.go
+++ b/internal/services/frontend/frontend.go
@@ -55,8 +55,8 @@ var (
 	server Server
 )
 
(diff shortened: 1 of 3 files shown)


internal/services/frontend/frontend.go
```diff
@@ -51,14 +55,14 @@ var (
 	server Server
 )
 
-// normalize the URLs we process before handing them off to the normal route
-// processing.
+// normalizeURL standardizes the URL string before handing them it to the
```
MikeCJohnsonUk, line 58: Extra "it" ?

PR #119, opened 2020-08-19
linter/inspection changes

Jim3Things:
Limited to store.go/storeapi.go

Did not address unused functions/symbols, as some functions still to be written.
Did not address exported symbols that are not used outside of store itself.  That needs further review.

### DIFF
--- a/internal/clients/store/store.go
+++ b/internal/clients/store/store.go
@@ -67,23 +67,23 @@
 package store
 
 import (
-    "context"
-    "fmt"
-    "log"
-    "strings"
-    "sync"
-    "time"
+	"context"
+	"fmt"
+	"log"
+	"strings"
+	"sync"
+	"time"
 
-    "github.com/Jim3Things/CloudChamber/internal/config"
-    "github.com/Jim3Things/CloudChamber/internal/tracing"
-    st "github.com/Jim3Things/CloudChamber/internal/tracing/server"
+	"github.com/Jim3Things/CloudChamber/internal/config"
+	"github.com/Jim3Things/CloudChamber/internal/tracing"
+	st "github.com/Jim3Things/CloudChamber/internal/tracing/server"
 
-    "go.etcd.io/etcd/clientv3"
-    "go.etcd.io/etcd/clientv3/concurrency"
-    ns "go.etcd.io/etcd/clientv3/namespace"
-    "go.etcd.io/etcd/etcdserver/api/v3rpc/rpctypes"
-    "google.golang.org/grpc/codes"
-    "google.golang.org/grpc/status"
+	"go.etcd.io/etcd/clientv3"
+	"go.etcd.io/etcd/clientv3/concurrency"
+	ns "go.etcd.io/etcd/clientv3/namespace"
+	"go.etcd.io/etcd/etcdserver/api/v3rpc/rpctypes"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 )
 
 // All CloudChamber key-value pairs are stored under a common namespace root.
@@ -94,9 +94,9 @@ import (
 // At no time can the root namespace be removed.
 //
 const (
-    slash                 = "/"
-    cloudChamberNamespace = string("/CloudChamber/v0.1")
-    testNamespaceSuffix   = string("Test")
+	slash                 = "/"
+	cloudChamberNamespace = string("/CloudChamber/v0.1")
+	testNamespaceSuffix   = string("Test")
 )
 
 // TraceFlags is the type used when setting of fetching the relevant flags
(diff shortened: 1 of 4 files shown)


internal/clients/store/store.go
```diff
@@ -225,9 +225,11 @@ func (store *Store) disconnected(ctx context.Context) error {
 // the back-end db service.
 //
 func Initialize(cfg *config.GlobalConfig) {
+	_ = st.WithSpan(context.Background(), tracing.MethodName(1), func(ctx context.Context) (err error) {
+		storeRoot.DefaultEndpoints = []string {
+			fmt.Sprintf("%s:%d", cfg.Store.EtcdService.Hostname, cfg.Store.EtcdService.Port),
```
MikeCJohnsonUk, line 230: It occurs to me, do we ever check anywhere on any of the "EP" values that port <= 65535? I didn't see anything in the settings code itself where EP->Port is just an int. Can viper handle an int16 field do you think?

I think that originally, I had port be an int16 type hence the %v in the fmt() rather than the %d.

internal/clients/store/store.go
```diff
@@ -1318,29 +1344,36 @@ func chkConditions(stm concurrency.STM, req *Request) error {
...
-// It is not an error to attmept to retrieve an empty set. For example, when querying
+// It is not an error to attempt to retrieve an empty set. For example, when querying
 // for the presence of a set of values, this method can be used which would successfully
 // return an empty set of key/value pairs if there are no matches for the supplied key
 // prefix.
 //
 func (store *Store) ListWithPrefix(ctx context.Context, keyPrefix string) (response *Response, err error) {
-
 	err = st.WithSpan(ctx, tracing.MethodName(1), func(ctx context.Context) (err error) {
+		var getResponse *clientv3.GetResponse
```
MikeCJohnsonUk, line 1354: ??

internal/clients/store/store.go
```diff
@@ -1406,7 +1440,9 @@ func (store *Store) ReadMultipleTxn(ctx context.Context, keySet RecordKeySet) (*
 			return nil
 		}
 
-		response, err := concurrency.NewSTM(
+		var response *clientv3.TxnResponse
```
MikeCJohnsonUk, line 1443: Again, don't understand the reason for this?

internal/clients/store/store.go
```diff
@@ -1638,17 +1675,15 @@ func (store *Store) ReadTxn(ctx context.Context, request *Request) (response *Re
 		return nil
 	})
 
-	return response, nil
+	return response, err
```
MikeCJohnsonUk, line 1678: Oops!

internal/clients/store/storeapi.go
```diff
@@ -449,40 +486,40 @@ func (store *Store) DeleteNew(ctx context.Context, r KeyRoot, n string, rev int6
 // NOTE: This should only be used at present if the number of user records
 //       is limited as there is a limit to the number of records that can
 //       be fetched from the store in a single shot. Eventually this will
-//       be updated to user an "interupted" enum style call to allow for
+//       be updated to user an "interrupted" enum style call to allow for
```
MikeCJohnsonUk, line 489: typo (in original) "...to use ..."

internal/clients/store/storeapi.go
```diff
@@ -449,40 +486,40 @@ func (store *Store) DeleteNew(ctx context.Context, r KeyRoot, n string, rev int6
...
+		response, err = store.ListWithPrefix(ctx, prefix)
 
 		if err != nil {
 			return err
 		}
 
 		recs := make(map[string]Record, len(response.Records))
 
-		for k, r := range response.Records {
+		for k, record := range response.Records {
```
MikeCJohnsonUk, line 512: So, it's happy with the "k" but not the "r" ?

internal/clients/store/store.go
```diff
@@ -297,7 +299,7 @@ func PrepareTestNamespace(ctx context.Context, cfg *config.GlobalConfig) {
 		if err := cleanNamespace(testNamespace); err != nil {
 			msg := fmt.Sprintf("failed to pre-clean the store as requested - namespace: %s err: %v", testNamespace, err)
 
-			st.Errorf(ctx, -1, msg)
+			_ = st.Errorf(ctx, -1, msg)
```
MikeCJohnsonUk, line 302: So, I'm curious, as a symmetry question, why does st.ErrorX() return an error which is typically either returned or discarded, but st.InfoX() does not? Should st.InfoX() perhaps return a nil error? Or alternatively, st.ErrorX() return nothing?

Of course, that does beg the question is it guaranteed that the ErrorX() function(s) will only ever return the supplied err parameter and never internally generate an error of their own?

internal/clients/store/store.go
```diff
@@ -370,19 +377,35 @@ func New(endpoints []string, timeoutConnect time.Duration, timeoutRequest time.D
...
-func (store *Store) Initialize(endpoints []string, timeoutConnect time.Duration, timeoutRequest time.Duration, traceFlags TraceFlags, namespaceSuffix string) error {
-
+func (store *Store) Initialize(
+		endpoints []string,
+		timeoutConnect time.Duration,
+		timeoutRequest time.Duration,
+		traceFlags TraceFlags,
+		namespaceSuffix string) error {
 	return st.WithSpan(context.Background(), tracing.MethodName(1), func(ctx context.Context) (err error) {
+		if err = store.connected(ctx); err != nil {
```
MikeCJohnsonUk, line 387: In other sequence like this, we've been using "err := xxx"

internal/clients/store/store.go
MikeCJohnsonUk: I tried changing the EP defn to be a uint16 for port which resulted in a pair of changes elsewhere, but otherwise it seems to be ok.

Jim3Things: Cool - let's go with the uint16.

internal/clients/store/store.go
Jim3Things: The response var declarations were done to remove the := in the call.  I was trying to avoid redeclaring err.

internal/clients/store/store.go
Jim3Things: see above

internal/clients/store/storeapi.go
Jim3Things: Yeah, 'r' is declared as an incoming parameter in the function, so k, r := redeclares r.

internal/clients/store/store.go
Jim3Things: Error (not Errorf) returns the error it is passed.  Errorf returns an error constructed from fmt.Errorf.  Both were done to enable the idiom of trace and return using 'return Error[f]'.  Let me go back through and see where I can move to return st.Error(..., error-value)

internal/clients/store/store.go
Jim3Things: It depends on whether or not we declare err as the return type name.  If we don't, and for the simple cases we don't need to, then ':=' works.  For the more complex cases, where we need an err variable, '=' should be more appropriate.

internal/clients/store/store.go
Jim3Things: I've looked just now and it seems that store always prepends some string in the trace entry.  If there are cases where that makes sense in the error itself, then those can move to st.Error.

internal/clients/store/store.go
MikeCJohnsonUk: Ok, I'll do that as a separate checkin.

MikeCJohnsonUk: Issue #120 filed.

internal/clients/store/store.go
MikeCJohnsonUk: But it won't re-declare err. The ":=" would only declare the getResponse var and continue to use the existing err variable - i.e. ":=" will only declare any currently undeclared variables in the assignment.

internal/clients/store/storeapi.go
MikeCJohnsonUk: Well huh, so it does.

internal/clients/store/store.go
MikeCJohnsonUk: I've been doing a little research. The underlying reason is that I'm not really a fan of the explicit discarding of a single response in the st.Errorf() case. If I were to refine this, I think my real disturbance in this particular case is that I don't see what the caller of the st.Errorf() function is supposed to do with any returned error. It doesn't seem pertinent to the overall code flow the caller is doing so if the underlying fmt.Xxx() failed then a panic() might be better? Or perhaps not. Not sure.

In the store[api] case, it may well be possible to fold the text into the error itself and move to st.Error(). I'll take a look next time I'm in there.

internal/clients/store/store.go
MikeCJohnsonUk: Ok, looking at the language spec, technically it will re-declare the err, but it must be of the same type and it does not introduce a new variable, just assigns a new value to the original.

I must say, I definitely prefer the := here to avoid the separate var declaration for the other variable. Is this something the linet complained about?

Jim3Things: Yeah, at least one did complain.  Thinking...

internal/clients/store/store.go
Jim3Things: It might make best sense to have a st.Fatalf, which replaces the log.Fatalf function, like we have with log.Errorf.  I'll make that change, and that would fix the issue, I believe.

internal/clients/store/store.go
Jim3Things: Ok, I'm going to put a bunch of them back.  The linters are inconsistently flagging this.  Sometimes they do, and sometimes they don't.

internal/clients/store/store.go
Jim3Things: Btw, the alternative would be: log.Fatal(st.Errorf(...args...))

Avoids creating the message and calling fmt.Sprintf.

Preferences?

PR #126, opened 2020-08-23
Add source file dependencies

MikeCJohnsonUk:
Fixes issue where a change to a single source file is insufficient for make to detect a rebuild is required.

Adds dependency chains based on major components and/or directories, so now a change to a single individual source file will be detected whenever a relevant make target is invoked.

We may choose to break this up or re-organize later on as this does look a little pedantic but at least it should work and be safe.

I'd welcome opinions on the matter.

ps. This is the second in the chain of PR's I mentioned in mail. There is no actual dependency on the Errorf() edits in any way.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,6 +40,128 @@ PROTO_GEN_FILES = \
     pkg/protos/trace_sink/trace_sink.pb.go
 
 
+SRC_CONFIG = \
+	internal/config/settings.go
+
+SRC_FRONTEND = \
+	$(SRC_CONFIG) \
+	$(SRC_STORE) \
+	$(SRC_TIMESTAMP) \
+	$(SRC_TRACING) \
+	$(SRC_TRACING_CLIENT) \
+	$(SRC_TRACING_SERVER) \
+	internal/services/frontend/DBInventory.go \
+	internal/services/frontend/DBUsers.go \
+	internal/services/frontend/errors.go \
+	internal/services/frontend/frontend.go \
+	internal/services/frontend/inventory.go \
+	internal/services/frontend/ping.go \
+	internal/services/frontend/session_manager.go \
+	internal/services/frontend/stepper.go \
+	internal/services/frontend/users.go \
+	internal/services/frontend/workloads.go
+
+SRC_MONITOR = \
+	internal/services/monitor/monitor.go
+
+SRC_SM = \
+	$(SRC_TRACING_SERVER) \
+	internal/sm/sm.go
+
+SRC_STEPPER_ACTOR = \
+    $(SRC_SM) \
+	$(SRC_TRACING) \
+	$(SRC_TRACING_SERVER) \
+	internal/services/stepper_actor/actor.go \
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -40,6 +40,128 @@ PROTO_GEN_FILES = \
...
+	internal/services/frontend/DBInventory.go \
+	internal/services/frontend/DBUsers.go \
+	internal/services/frontend/errors.go \
+	internal/services/frontend/frontend.go \
+	internal/services/frontend/inventory.go \
+	internal/services/frontend/ping.go \
+	internal/services/frontend/session_manager.go \
+	internal/services/frontend/stepper.go \
+	internal/services/frontend/users.go \
+	internal/services/frontend/workloads.go
```
Jim3Things, line 62: I think this is a general comment - can you dynamically build these up for each directory?  Use a for loop to scan for .go files in the appropriate directory that don't end in "_test.go" and use that to construct a directory-level symbol, and then assemble the dependencies from that?

I ask because I think it is going to be hard to remember to add each new .go file here as we create them.

Makefile
```diff
@@ -40,6 +40,128 @@ PROTO_GEN_FILES = \
...
+
+SRC_MONITOR = \
+	internal/services/monitor/monitor.go
+
+SRC_SM = \
+	$(SRC_TRACING_SERVER) \
+	internal/sm/sm.go
+
+SRC_STEPPER_ACTOR = \
+    $(SRC_SM) \
```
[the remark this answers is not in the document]
MikeCJohnsonUk: Sigh. Silent spaces...

Fixed.

Makefile
MikeCJohnsonUk: This is where I expected the comments.

Normally, for single-directory based makefiles I would anticipate a wildcard dependency along the lines of *.go, but because Go believes that test code should be in <module>_test.go, the wildcard would result in matching the test code for the production build.

It may be possible to use something a little more selective than a simple wildcard, but I need to do some research on this one. The current layout at least has the advantage of working (at present) even though it is somewhat kludgy and I wanted to get something in asap.

We also really have a similar issue with the proto files. Something else I need to think on. Unfortunately changes here may need a file location change assuming there is no placement requirement inherent in the protobuf compilation?

Jim3Things: I worry a little less about the protobuf files, and we can place them where it makes sense - although that may cause issues with the package names, not sure.

I'm ok with going with this for now, so long as we do the investigation to see how to get away from this list in a later checkin.

MikeCJohnsonUk: Updated to user a user defined function which makes use of the builtin $(filter-out, xxx) and $(wildcard, xxx) functions to generate a list of all the .go files which do not have and _test suffix in the file name portion.

I've done some testing to satisfy myself that updated regular source files trigger a build, but update test files do not, without affecting the "test" target which works as before.

I think this helps as it avoid listing all the source files explicitly, but let me know what you think.

PR #131, opened 2020-08-25
Stepper bugfix, minor cleanup

Jim3Things:
While looking to see if I could speed up the sim_supportd startup, I found an issue in stepper.  I've also added a banner to the traces that shows when the execution started.

### DIFF
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -11,6 +11,8 @@ import (
 
 	"go.opentelemetry.io/otel/api/global"
 	"go.opentelemetry.io/otel/api/trace"
+
+	"github.com/Jim3Things/CloudChamber/internal/tracing"
 )
 
 //go:generate go run generator/generate.go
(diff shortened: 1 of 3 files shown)


pkg/version/version.go
```diff
@@ -42,14 +44,21 @@ func Show() {
...
 
-// TraceVersion is a simple function to insert a trace record containing the
-// version information into the trace stream.
+// Trace is a simple function to insert the leading trace events that show the
+// startup of the service, and the version information associated with it.
 //
 func Trace() {
 	tr := global.TraceProvider().Tracer("")
 
 	_ = tr.WithSpan(context.Background(), "TraceVersion", func(ctx context.Context) (err error) {
```
MikeCJohnsonUk, line 53: Clearly I renamed the function at some point, but missed stuff. Is the tracing.MethodName(1) function available here?

Jim3Things: Yes, it is.  Added.

PR #133, opened 2020-08-26
Change console output to be re-directed to a log file

MikeCJohnsonUk:
When starting the service executables, redirect any console output (stdout and stderr) to a log file instead.

### DIFF
--- a/scripts/StartCloudChamber.cmd
+++ b/scripts/StartCloudChamber.cmd
@@ -13,7 +13,7 @@ set SCRIPTDIR=%~dp0
 set CLOUDCHAMBERDIR=%SCRIPTDIR:~0,-7%
 set CLOUDCHAMBERFILE=%CLOUDCHAMBERDIR%\Files
 set CLOUDCHAMBERDATA=%CLOUDCHAMBERDIR%\Data
-
+set CLOUDCHAMBERLOGS=%CLOUDCHAMBERDIR%\Logs
 
 
 rem Check for requests for help without actually doing anything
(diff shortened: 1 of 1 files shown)


scripts/StartCloudChamber.cmd
```diff
@@ -82,7 +108,7 @@ rem
 echo.
 echo Starting %TARGETBIN%
 
-start %TARGETBIN% -config=%2
+start %TARGETBIN% -config=%2 2>&1 >%CLOUDCHAMBERLOGS%\%BINARY:~0,-4%-%UpdateDateTime%.log
```
Jim3Things, line 111: I'm going to assume this is right, but I'll confess that I'm not sure what the file name will end up being.

MikeCJohnsonUk: For a deployment based at C:\CloudChamber, the default, you would end up with something like

C:\Chamber\src\github.com\Jim3Things\CloudChamber>dir c:\CloudChamber\Logs
 Volume in drive C has no label.
 Volume Serial Number is 0C24-9DA7

 Directory of c:\CloudChamber\Logs

08/25/2020  04:56 PM    <DIR>          .
08/25/2020  04:56 PM    <DIR>          ..
08/25/2020  04:56 PM                 0 controllerd-20200825-165638.log
08/25/2020  04:56 PM                 0 inventoryd-20200825-165638.log
08/25/2020  04:56 PM                 0 sim_supportd-20200825-165638.log
08/25/2020  04:56 PM                 0 web_server-20200825-165638.log
               4 File(s)              0 bytes
               2 Dir(s)  219,362,406,400 bytes free

PR #136, opened 2020-08-28
Add REST service for trace_sink

Jim3Things:
- Add log entry retrieval to frontend
- Add trace_sink client
- Add Stringer interface to Endpoint
- Add more general error handling for numeric values in query fields
- Add infrastructure client interceptor for grpc
- Fix deadlock in trace_sink

### DIFF
--- a/internal/services/frontend/common_test.go
+++ b/internal/services/frontend/common_test.go
@@ -82,7 +82,7 @@ func commonSetup() {
 	}
 
 	if err := tracing_sink.Register(s); err != nil {
-		log.Fatalf("Failed to register tracking sink: %v", err)
+		log.Fatalf("Failed to register tracing sink: %v", err)
 	}
 
 	go func() {
--- a/internal/services/frontend/common_test.go
+++ b/internal/services/frontend/common_test.go
@@ -27,8 +27,10 @@ import (
 	"google.golang.org/grpc/test/bufconn"
 
 	ts "github.com/Jim3Things/CloudChamber/internal/clients/timestamp"
+	tsc "github.com/Jim3Things/CloudChamber/internal/clients/trace_sink"
 	"github.com/Jim3Things/CloudChamber/internal/config"
 	stepper "github.com/Jim3Things/CloudChamber/internal/services/stepper_actor"
+	"github.com/Jim3Things/CloudChamber/internal/services/tracing_sink"
 	ctrc "github.com/Jim3Things/CloudChamber/internal/tracing/client"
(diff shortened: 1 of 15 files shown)


internal/services/tracing_sink/sink.go
```diff
@@ -149,7 +149,12 @@ func (s *server) GetAfter(ctx context.Context, request *pb.GetAfterRequest) (*pb
 			return resp.err
 		}
 
-		resp = <- s.wait(id, request.MaxEntries)
+		ch := s.wait(id, request.MaxEntries)
+
+		s.mutex.Unlock()
+		resp = <- ch
+		s.mutex.Lock()
```
MikeCJohnsonUk, line 156: Surely you expected a comment from me here right? Looking at the need for the lock, as far as I can see it is only required for lines 147 to 152. Would it be better to just acquire it to cover that region rather than this unlock() / re-lock() pair where we only have the re-lock() to satisfy the defer in the outer context where we don't even need the lock anyway?

internal/services/frontend/logs.go
```diff
@@ -1,22 +1,91 @@
...
+			ctx, w, r,
+			func(ctx context.Context, session *sessions.Session) error {
+				return ensureEstablishedSession(session)
+			})
+
+		if err != nil {
+			return httpError(ctx, w, err)
+		}
+
+		fromId, err := ensureNumber("from", from)
```
MikeCJohnsonUk, line 42: Are we allowing "from" to be negative to perhaps indicate a relative offset?

Jim3Things: We do allow it to be negative.  From is logically the last entry we've processed, so -1 is just before the first possible entry.

internal/services/frontend/common_test.go
```diff
@@ -79,6 +81,10 @@ func commonSetup() {
 		log.Fatalf("Failed to register stepper actor: %v", err)
 	}
 
+	if err := tracing_sink.Register(s); err != nil {
+		log.Fatalf("Failed to register tracking sink: %v", err)
```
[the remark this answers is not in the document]
Jim3Things: fixed

internal/services/tracing_sink/sink.go
Jim3Things: Yeah, I tried too hard to keep the defer pattern.

internal/services/frontend/logs.go
MikeCJohnsonUk: Ok

PR #137, opened 2020-08-29
Minor fixes

Jim3Things:
- Use a test yaml file for frontend tests, rather than hand-crafted struct
- Use channel timeouts for test delays, put into common package
- Move setting the return cookies on stepper wait-for until after the wait completes
- Add sink tests to makefile test target

### DIFF
--- a/internal/services/frontend/logs_test.go
+++ b/internal/services/frontend/logs_test.go
@@ -11,6 +11,7 @@ import (
 	"github.com/stretchr/testify/require"
 
 	tsc "github.com/Jim3Things/CloudChamber/internal/clients/trace_sink"
+	"github.com/Jim3Things/CloudChamber/internal/common/channels"
 	"github.com/Jim3Things/CloudChamber/internal/tracing/exporters/unit_test"
 	"github.com/Jim3Things/CloudChamber/pkg/protos/log"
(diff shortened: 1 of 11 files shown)


internal/services/frontend/logs_test.go
```diff
@@ -36,19 +37,23 @@ func testLogsGetPolicy(t *testing.T, cookies []*http.Cookie) (*pb.GetPolicyRespo
...
-	res := &pb.GetAfterResponse{}
-	err := getJSONBody(response, res)
+		res := &pb.GetAfterResponse{}
+		err := getJSONBody(response, res)
 
-	assert.Nilf(t, err, "Unexpected error, err: %v", err)
+		assert.Nilf(t, err, "Unexpected error, err: %v", err)
 
-	return res, response.Cookies()
+		return res, response.Cookies()
```
MikeCJohnsonUk, line 56: Has something gone screwy with the indentation again? These look to have had a second tab inserted?

Jim3Things: No, it was intentional.  The function declaration line was too long, so I changed the line to have the parameters stacked.  I could either double indent the parameters or the body.  Following from some previous code format fixes, I double indented the body.

PR #138, opened 2020-08-30
Move grpc protos into a common subdirectory

Jim3Things:
.. and move the request and response messages into a common proto file to allow for sharing.

Note that when this checkin is complete the subdirectories pkg/protos/Stepper, pkg/protos/Monitor, and pkg/protos/trace_sink will need to be deleted.


pkg/protos/services/requests.proto
```diff
@@ -125,3 +100,66 @@ message StatusResponse {
...
+// Return the active policies for the trace sink service
+message GetPolicyResponse {
+    // The limit on the number of entries held in the trace sink
+    int64 max_entries_held = 1;
+
+    // Earliest trace entry currently held.
+    int64 first_id = 2;
+}
+
+// --- trace sink specific messages
```
MikeCJohnsonUk, line 165: For the most part, I understand the edit, but what I was a little unclear on was why there messages are in the "requests.proto" file rather than the "trace_sink.proto" file given they are labeled as track_sink specific?

Jim3Things: I started with the requests in the specific proto files, but the messages all share the same package name, so there are name collisions to deal with (e.g. ResetRequest).  I could have only put the collision into a common file, but I thought it was more stable and obvious to put all the messages into the same file so that collisions would be obvious more quickly.

PR #143, opened 2020-09-07
Old API set removal from the store package

MikeCJohnsonUk:
The goal is to remove a bunch of obsolete, but not quite dead code implementing an old API set to the store.

Background

When I was originally implementing the store, I built a number of routines Read(), Write() etc, a number of expansions on those calls, along with all the tests to verify them.

Later, I discovered the concurrency etcd package which made transactional use of etcd pretty easy and so added a set of transactional calls ReadTxn(), WriteTxn() and DeleteTxn() which superseded the original API set. I also added some tests and this is where I made the first mistake: almost all of the new tests made extensive use of the original API set to prepare test state and then verify it afterwards.

Later still, when I added the StoreAPI.go module to interface the store transactional code to the DbUsers.go module, I needed to name the simplified API routines but found all the names I wanted to use conflicted with the (mostly) redundant original API set and so ended up with names of the form ReadNew(), DeleteNew() and so on.

This was getting ridiculous.

This Change

What this PR is about is cleaning all this up. Specifically this PR:-
- removes all of the original API set (~600 lines)
- removes structures used by the original API set (RecordUpdateSet, KeyValueArg, KeyValueResponse)
- removes all of the original API usage in the test code
- adds some missing test coverage for the XxxxTxn() routines, in some cases for features we were already using

The Consequences

There is relatively little change in the store.go production code.
The old API purge has effectively lead to almost all the test code being re-written.

The names in the StoreAPI module will be sorted out in a follow-on PR.

Also clears the way for some more complicated operations such as allowing transactional read/modify/write within a single transaction.


internal/clients/store/common_test.go
```diff
@@ -40,58 +45,255 @@ func commonSetup() {
...
-	for _, kv := range keyValueSet {
-		recordUpdateSet.Records[kv.key] =
-			RecordUpdate{
-				Condition: condition,
-				Record: Record{
-					Revision: RevisionInvalid,
-					Value:    kv.value,
-				},
-			}
+func testGenerateRequestFromWReadResponseWithCondition(response *Response, condition Condition) *Request {
```
Jim3Things, line 197: FromWRead?  Should it be FromRead?

internal/clients/store/common_test.go
```diff
@@ -40,58 +45,255 @@ func commonSetup() {
...
+	}
+
+	// Finally, do we have a write request/response pair for each
+	// record in the read response
+	//
+	for k, r := range readResponse.Records {
+
+		wReq, ok := writeRequest.Records[k]
+
+		require.Truef(t, ok, "No write request record to match read response record")
```
Jim3Things, line 283: It's a nit, as I read through the tests I see cases like this and I think, but am not sure, that they don't need the '...f' form of the call - as they only have a constant string to emit.

internal/clients/store/store.go
```diff
@@ -1387,184 +926,39 @@ func (store *Store) ListWithPrefix(ctx context.Context, keyPrefix string) (respo
...
-			return nil
-		}
-
-		response, err := concurrency.NewSTM(
-			store.Client,
-			actionWriteRecords,
-			concurrency.WithIsolation(concurrency.Serializable),
-			concurrency.WithPrefetch(*prefetchKeys...),
-		)
+		opCtx, cancel := context.WithTimeout(context.Background(), store.TimeoutRequest)
```
Jim3Things, line 938: Why do you create a new context that is unconnected to the caller's context here?

internal/clients/store/common_test.go
MikeCJohnsonUk: Yes. Fixed

internal/clients/store/store.go
MikeCJohnsonUk: This one actually predates our using contexts!

This is used as part of the construction of the cancel context. It's how the examples did it. I've no idea whether or not we can connect them to the passed in context and if so what the benefit might be. Should we wish to change this I presume this would end up along the lines of 

		opCtx, cancel := context.WithTimeout(ctx, store.TimeoutRequest)
		opResponse, err := store.Client.Delete(opCtx, keyPrefix, clientv3.WithPrefix())
		cancel()

Is this safe?

internal/clients/store/common_test.go
MikeCJohnsonUk: I'm a little in two minds over this. In many cases now, I've started using the non-Xxxxf() form given that the expected and actual value is part of the emitted message. In some cases though, I would prefer to have a message as a quick indicator of what the failed assert signifies, regardless of the need to print specific values. That could also be achieved via a comment in the source, but a message seems more likely to remain applicable. The problem without a specific message is that although not strictly necessary, there is sometimes a need for interpretation of the source.

However, I'm sure there are many where I was just automatically adding text, 'cos that was what I'd been doing before.

Actually, I've just been scanning some of these and it looks like there are quite a lot of cases where the message and params could just be flat out removed.

I think I want to deal with this in a separate pass. I've added issue #145  to cover this.

Jim3Things: That's fine with me.

internal/clients/store/store.go
Jim3Things: AFAIK, it should be.

Jim3Things (approved):

PR #144, opened 2020-09-07
Simplify tracing code

Jim3Things:
The trace exporters are rewritten into a simpler structure.
Deferrable now obsolete and removed.
Global state in the exporters is now gone.

There is a mechanical test structure change throughout

Makefile simplified


internal/config/InventoryReader_test.go
```diff
@@ -24,27 +28,28 @@ func TestMain(m *testing.M) {
 // Establish the test environment, including starting a test frontend service
 // over a faked http connection.
 func commonSetup() {
-	setup.Init(exporters.UnitTest)
+	utf = exporters.NewExporter(exporters.NewUTForwarder())
+	exporters.Init(utf)
```
MikeCJohnsonUk, line 32: This seems a little, ah, odd? I would have expected that NewExporter() would return an initialized utf. Init() seems to be for connecting (possibly multiple) exporters to a single trace provider?

internal/tracing/exporters/config.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"log"
+
+	"go.opentelemetry.io/otel/api/global"
+	sdk "go.opentelemetry.io/otel/sdk/trace"
+)
+
+// Init configures one or more OpenTelemetry exporters into our trace provider
+func Init(exporters ...*Exporter) {
```
MikeCJohnsonUk, line 11: Following on from the other comment, perhaps this should be "ConnectToProvider()" or something along those lines. I think my earlier confusion was expecting some sort of initialization action on the exporter, but that's not really what's happening?

Jim3Things: Good idea.  I'll use that.

MikeCJohnsonUk (approved):

PR #149, opened 2020-09-12
Simplify span handling, add infrastructure inheritence

Jim3Things:
This adds a wrapped span that is used to inherit the infrastructure attribute on spans, it adds a non-WithSpan sequence, modifies users.go & DBUsers.go to use that, with some cleanup of the http error handling.

The new start span, and the existing trace event methods, are moved to a common tracing level.  The client and server subdirectories will move to contain client and service interceptors, as the WithSpan variants are removed.

### DIFF
--- a/internal/services/frontend/DBUsers.go
+++ b/internal/services/frontend/DBUsers.go
@@ -17,8 +17,9 @@ import (
 	"golang.org/x/crypto/bcrypt"
 
 	"github.com/Jim3Things/CloudChamber/internal/clients/store"
+	"github.com/Jim3Things/CloudChamber/internal/common"
 	"github.com/Jim3Things/CloudChamber/internal/config"
-	st "github.com/Jim3Things/CloudChamber/internal/tracing/server"
+	"github.com/Jim3Things/CloudChamber/internal/tracing"
 	pb "github.com/Jim3Things/CloudChamber/pkg/protos/admin"
 )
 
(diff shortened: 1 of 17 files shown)


internal/services/frontend/DBUsers.go
```diff
@@ -31,62 +32,66 @@ type DBUsers struct {
...
-					"CloudChamber: standard %q account is not using using configured password - error %v",
-					cfg.WebServer.SystemAccount, err)
-			}
-
-			return nil
+		if err = bcrypt.CompareHashAndPassword(
+			existingUser.GetPasswordHash(),
+			[]byte(cfg.WebServer.SystemAccountPassword)); err != nil {
+			tracing.Infof(
+				ctx, tick,
```
MikeCJohnsonUk, line 88: Do I remember hearing you saying something about storing the current "tick" value in the ctx or am I getting mixed up with something else?

Is that something we could do, letting the trace code just extract the value and then update the ctx value at the appropriate time? 

Just a notion, not even a suggestion.

internal/tracing/spanEx.go
```diff
@@ -0,0 +1,204 @@
...
+	"go.opentelemetry.io/otel/api/global"
+	"go.opentelemetry.io/otel/api/kv"
+	"go.opentelemetry.io/otel/api/trace"
+
+	"github.com/Jim3Things/CloudChamber/pkg/protos/log"
+)
+
+// spanEx extends the OpenTelemetry Span structure with additional features
+// needed by CloudChamber
+type spanEx struct {
```
MikeCJohnsonUk, line 18: Does this portend spanExEx and/or spaceEx2?

Sorry, couldn't resist ;-)

internal/services/frontend/DBUsers.go
Jim3Things: That's what I think we'll end up doing -- the discussion about storing the tick in the context is a future item though.

internal/tracing/spanEx.go
Jim3Things: Open to other names.  Just couldn't think of one that wouldn't be confusing or really long...

internal/services/frontend/DBUsers.go
MikeCJohnsonUk: Fair enough

internal/tracing/spanEx.go
MikeCJohnsonUk: Just yanking your chain right now. I've no suggestions.

MikeCJohnsonUk (approved):

PR #152, opened 2020-09-14
Move frontend, tracing sink away from WithSpan

Jim3Things:
... plus some minor support utility cleanup

### DIFF
--- a/internal/services/tracing_sink/sink.go
+++ b/internal/services/tracing_sink/sink.go
@@ -13,9 +13,8 @@ import (
 	"github.com/golang/protobuf/ptypes/empty"
 	"google.golang.org/grpc"
 
-	"github.com/Jim3Things/CloudChamber/internal/common"
-    "github.com/Jim3Things/CloudChamber/internal/tracing"
-    st "github.com/Jim3Things/CloudChamber/internal/tracing/server"
+	"github.com/Jim3Things/CloudChamber/internal/clients/timestamp"
+	"github.com/Jim3Things/CloudChamber/internal/tracing"
 	"github.com/Jim3Things/CloudChamber/pkg/protos/log"
 	pb "github.com/Jim3Things/CloudChamber/pkg/protos/services"
 )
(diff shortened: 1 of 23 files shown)


internal/services/tracing_sink/sink.go
```diff
@@ -94,82 +93,85 @@ func Register(svc *grpc.Server) error {
...
-	err := st.WithInfraSpan(ctx, func(ctx context.Context) error {
-		if err := request.Validate(); err != nil {
-			return err
+	var err error = nil
+
+	ctx, span := tracing.StartSpan(context.Background(),
+		tracing.AsInternal())
+
+	// Pick up the current time to avoid repeatedly fetching the same value
+	tick := clients.Tick(ctx)
```
MikeCJohnsonUk, line 137: Is the current tick something we need to collect at the beginning of the span, or could it in fact be moved to within the deferred func()?

internal/services/tracing_sink/sink.go
```diff
@@ -180,21 +182,24 @@ func (s *server) GetPolicy(ctx context.Context, _ *pb.GetPolicyRequest) (*pb.Get
...
+	ctx, span := tracing.StartSpan(context.Background(),
+		tracing.AsInternal())
+	defer span.End()
 
-		tracing.Infof(
-			ctx, common.Tick(ctx),
-			"GetPolicy returning; firstId=%d, maxEntriesHeld=%d",
-			resp.FirstId, resp.MaxEntriesHeld)
+	// Pick up the current time to avoid repeatedly fetching the same value
+	tick := clients.Tick(ctx)
```
MikeCJohnsonUk, line 190: I guess a similar question to that above. Is this here because we want to fetch it before ?

internal/services/tracing_sink/sink.go
```diff
@@ -94,82 +93,85 @@ func Register(svc *grpc.Server) error {
 }
 
 // Append adds a trace entry to the list of known entries.
-func (s *server) Append(ctx context.Context, request *pb.AppendRequest) (*empty.Empty, error) {
+func (s *server) Append(_ context.Context, request *pb.AppendRequest) (*empty.Empty, error) {
```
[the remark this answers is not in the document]
Jim3Things: We don't really get a choice here - this is the signature that grpc comes up with.

internal/services/tracing_sink/sink.go
Jim3Things: We probably could here, and it would be more correct.  I'll make that change.

internal/services/tracing_sink/sink.go
Jim3Things: This is an intermediate form of the change that I want to make.  I'm not sure how to do it in one go that doesn't end up spanning most of our Go code.

My thought is for StartSpan to load the context with the current time; for calls that either advance the time or wait to also update the context; and for the trace calls to lose the tick argument.

This change is just trying to always get tick set up right after the StartSpan so that the implications of StartSpan setting the context are somewhat checked out before we make that change.

internal/services/tracing_sink/sink.go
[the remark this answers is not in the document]
MikeCJohnsonUk: Ok

internal/services/tracing_sink/sink.go
MikeCJohnsonUk: Fair enough. I like the sound of the end goal.

MikeCJohnsonUk (approved):

PR #154, opened 2020-09-17
Inventory to cloud chamber

manpreet85:
1. Updated Inventory.Yaml to represent 8 racks and each rack having 8 blades.
2. Updating DBInventory.go so that the Inventory is read from the external defintion file.


internal/services/frontend/DBInventory.go
```diff
@@ -167,11 +167,11 @@ func InitDBInventory() error {
 
 // GetMemoData returns the maximum number of blades held in any rack
 // in the inventory.
-func (m *DBInventory) GetMemoData() (int, int64, *common.BladeCapacity) {
+func (m *DBInventory) GetMemoData() (int64, *common.BladeCapacity) {
```
Jim3Things, line 170: The change to add this was made in the recent checkins, and inventory.go depends on it.  Was removing it intentional?

manpreet85: Unintentional. 

Should I reenter and send for PR

Jim3Things: Sure.

MikeCJohnsonUk: Looks like an older version? We're missing the "int" return value in the declaration, and the associated returned value on line 174

internal/services/frontend/DBInventory.go
```diff
@@ -277,3 +277,7 @@ func maxInt64(a int64, b int64) int64 {
 
 	return a
 }
+
+
+//Cloud chamber config file needs to have a setting that point to inventory.yaml file
+//We need dbinventory to create the inventory via the inventory reader
```
MikeCJohnsonUk, line 283: Do we need to retain these comments?

MikeCJohnsonUk (changes requested):

Jim3Things (approved):

MikeCJohnsonUk (approved):

PR #155, opened 2020-09-18
Add WithReason, enhance log entry validation checking

Jim3Things:


internal/services/frontend/frontend.go
```diff
@@ -94,13 +94,12 @@ func normalizeURL(next http.Handler) http.Handler {
 // arrives.  This is used, for example, to trace file access requests.
 func traceRequest(spanName string, next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		ctx, span := tracing.StartSpan(context.Background(),
-			tracing.WithName(spanName),
+		_, span := tracing.StartSpan(context.Background(),
+			tracing.WithName(fmt.Sprintf("%s: %s %q", spanName, r.Method, r.URL.String())),
```
MikeCJohnsonUk, line 98: Are we sure sufficient validation on the request has taken place by this point? Is suspect so, but I thought I'd ask.

internal/services/tracing_sink/sink_test.go
```diff
@@ -77,12 +77,47 @@ func commonSetup(t *testing.T) (context.Context, *grpc.ClientConn) {
 }
 
 func createEntry(events int) *log2.Entry {
-	tag := rand.Int()
+	tag := rand.Int63()
+	if tag == 0 {
+		tag = 1
+	}
```
MikeCJohnsonUk, line 83: Ahem? DRY?

internal/services/frontend/frontend.go
Jim3Things: I think so - we don't do any further validation, so this is as good as it is when we use it.

internal/services/tracing_sink/sink_test.go
Jim3Things: I'll be honest, I wasn't sure that I could always add one to tag.  I want a random number in the range of [1, maxint64).  Looking at the documentation, it looks like I can add 1.  So I'll change it to that.

MikeCJohnsonUk: Ok, crossed-purposes. 

I was talking about there being two (at least) instances of the if {} test and clause. Hence DRY.

The problem now is that if rand.Int63() returns 0x7fff...ffff then the +1 will overflow a max positive int64, so unless everything on the r.h.s. is cast/co-erced to a uint64 before the addition, there is potential for an overflow.

Unless int63() returns a signed 63bit int (i.e. 0 <= n ,<= 0x3fff...ffff. I'd call the docs a little ambiguous on that. If only they had specified the return value as being within an explicit range this would be so much clearer.

Jim3Things: Ah, ok.  I've refactored to remove both the ambiguity and the duplication. PTAL.

MikeCJohnsonUk (approved):

PR #158, opened 2020-10-03
DBInventoryActual-04Oct20

manpreet85:
Hi Jim and Mike, 

I am struggling with the test. System response for every instance of test is in following image

image
 Please advise


Jim3Things:
Hi Manpreet,

The error that you're seeing is that your test uses 'package main' rather than 'package frontend'.  Go doesn't allow a directory to have multiple packages like that.

There are two hints for this: the error says that it found packages frontend and main in the same directory, and the 'package main' line in the code has a problem squiggly underline - if you look at the problems tab it will likely tell you the same thing.  They're both saying that there is an inconsistent package name in the files.

HTH,
Jim.

internal/services/frontend/DBInventoryActual.go
```diff
@@ -0,0 +1,94 @@
+package frontend
```
Jim3Things, line 1: This is what we looked at last night, so is what I expected.  Do remember that before checking in you should do the pass to ensure the right comments and blank line structure.

internal/services/frontend/DBInventoryActual_test.go
```diff
@@ -0,0 +1,16 @@
+package main
+
+import (
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func TestInitDBInventoryActual(t *testing.T){
+
+	response, err := InitDBInventoryActual(inven *DBInventory)
```
Jim3Things, line 10: There are a couple of issues here.

First, the call to InitDBInventoryActual expects to receive a fully filled in DBInventory structure.  It looks like you're trying to declare a parameter here, which is what you do when defining the function (which you did in the module above).  Rather than that, you need to add code that reads in and constructs a variable with a *DBInventory type, and pass that variable through here.

(As a hint, since this is a frontend test, take a look at how the test environment is initialized)

Second, while it isn't yet an issue, you need to add the opening and closing of the unit test tracing channel like we do in the other tests.

Finally, since this is a function call and not an HTTP message, response is probably not a good variable name.  Think about what would be more accurate.

internal/services/frontend/DBInventoryActual_test.go
```diff
@@ -0,0 +1,16 @@
...
+	require.NotNil(t, err)
+	r, ok := actual.Zone.Racks["rack1"]
+	require.True(t, ok)
+	assert.Equal(t, 8, len(r.Blades))
```
Jim3Things, line 14: It's a start, but consider what you're trying to prove - that the resulting actual structure has the same key structure as the inventory definition, and that all status fields have the expected values.  How would you verify that?

MikeCJohnsonUk:
Further to Jim's comment, if you compare the start of InventoryActual_test.go to the start of Inventory_test.go you can see how the current test module declares the package name.

Also, wrt (with respect to) your picture, another clue to emphasize the squiggly underline is the "problem" tab showing there is (at least) one outstanding problem. 

When looking at the list of problems VsCode has detected, all the issues/defects marked with a red circled 'X' will need to be fixed to allow the file to compile. All the warnings marked with a '!' character in a yellow triangle are areas where there is something that's not quite right, but it should not prevent a compilation. Typically these warnings are a lack of a suitably formatted comment which are not of immediate concern, but probably should be fixed before the checking process is started.

Jim3Things (approved):

MikeCJohnsonUk (approved):

PR #159, opened 2020-10-04
Fix edge timing issue in forwarder.

Jim3Things:
If the goroutine stalls for long enough after issuing the close-ack it can find that the forwarder is open again and we can end up with two active forwarders.

### DIFF
--- a/internal/tracing/exporters/exporters.go
+++ b/internal/tracing/exporters/exporters.go
@@ -132,7 +132,7 @@ func (e *Exporter) Close() {
 func (e *Exporter) processLoop(ch chan interface{}) {
 	early := true
 
-	for !e.closed {
+	for {
 		msg := <-ch
 
 		switch pkt := msg.(type) {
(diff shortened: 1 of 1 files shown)


MikeCJohnsonUk (approved):
Whilst this is definitely a step in the right direction, I'd like to pose a few questions.

What exactly is the mutex in the exporter protecting? Right now that excludes the 'closed' flag.

Should the setting of the 'closed' flag be moved to the Close() routine itself? Alternatively, should the clearing of the 'closed' flag be moved to the processLoop() function?

What does the 'closed' flag actually represent? It almost seems as if it is trying to represent 3 states. opened, opened with process loop running, closed.

Jim3Things:
The mutex is ensuring that an open and close operations are serialized.  There's an invariant that there are only ever 0 or 1 active processloop goroutines at a time (there may be more goroutines, but those have reached a point where they are only waiting to be scheduled to exit), and the closed flag is supposed to track whether we're at 0 or 1.

The bug was that a processloop instance could end up changing its mind about whether or not it was exiting, which would then break the 0 or 1 invariant.

And, yes, looking at the current code, the closed flag could be set in the close function after receiving the close-ack message from the processloop.  That would be cleaner, and I'll make that change at some point.

PR #161, opened 2020-10-06
Backlog of tracing features

Jim3Things:
This change adds functions to update the span name and reason text.  It also removes the Xxxf tracing variants and enhancing the Xxx forms to do formatting if needed.   And the close flag on the forwarder loop is moved as per a previous review comment.

Finally, users.go has been updated to change the update the span names to something more specific once we know the username.


internal/tracing/spanEx.go
```diff
@@ -234,4 +237,22 @@ func logError(ctx context.Context, err error) error {
 	return err
 }
 
+// formatIf determines if this is a simple string, or something to format
+// before returning.
+func formatIf(a ...interface{}) string {
```
MikeCJohnsonUk, line 242: So you're guaranteed 'a' exists even if it has no entries?

internal/tracing/exporters/common.go
```diff
@@ -67,11 +68,22 @@ func extractEntry(_ context.Context, data *trace.SpanData) *log.Entry {
...
+		switch item.EventAction {
+		case log.Action_UpdateSpanName:
+			entry.Name = item.Text
+
+		case log.Action_UpdateReason:
+			entry.Reason = item.Text
+
+		default:
+			entry.Event = append(entry.Event, &item)
+		}
```
MikeCJohnsonUk, line 86: So updating either the snap name or the reason does not give rise to an entry in the output? That is, when viewing the trace, I would not see an entry resulting from the update itself?

Do you have a section of output to show how this actually appears?

internal/tracing/spanEx.go
```diff
@@ -131,90 +132,92 @@ func StartSpan(
...
+		}
 	}
 
-	panic("Invalid Error call - no valid arguments found")
-}
-
-// Errorf posts an error trace event with a complex string formatting
-func Errorf(ctx context.Context, f string, a ...interface{}) error {
-	return logError(ctx, fmt.Errorf(f, a...))
+	return logError(ctx, fmt.Errorf(a[0].(string), a[1:]))
```
MikeCJohnsonUk, line 215: And if len(a) should happen to be zero?

Should perhaps these all be func Xxx(ctx, fmt, a ...interface{}) to guarantee the presence of the fmt string?


Unless we actually have Info(), Warn(), Error() etc cases without a message string? Given that the '0' case is handled for at least the Info() and Warn() cases so I assume so.

internal/services/frontend/users.go
```diff
@@ -355,13 +361,15 @@ func handlerUserOperation(w http.ResponseWriter, r *http.Request) {
 		vars := mux.Vars(r)
 		username := vars["username"]
 
-		tracing.Infof(ctx, "Operation %q, user %q, session %v", op, username, session)
+		tracing.Info(ctx, "Operation %q, user %q, session %v", op, username, session)
```
MikeCJohnsonUk, line 364: Should this line be moved to the following "default" switch case? Of course, that pre-supposes that the UpdateSpanName() does include adding an entry to the trace.

internal/tracing/exporters/common.go
Jim3Things: That's correct, there is no trace entry kept - only the change to the span itself.

An example is output is:

[8566bbaa7a5ed71a:0000000000000000] (Infra) OK:  Logging out user "admin"():
    D:/Projects/go/src/github.com/Jim3Things/CloudChamber/internal/tracing/StackData.go:43
    D:/Projects/go/src/github.com/Jim3Things/CloudChamber/internal/tracing/spanEx.go:90
    D:/Projects/go/src/github.com/Jim3Things/CloudChamber/internal/services/frontend/users.go:351
    D:/Projects/go/go1.15.1/src/net/http/server.go:2042
    D:/Projects/go/src/github.com/Jim3Things/CloudChamber/vendor/github.com/gorilla/mux/mux.go:210

           : [I] () Called "/services.Stepper/Now", GRPC completion status: Success
        D:/Projects/go/src/github.com/Jim3Things/CloudChamber/internal/tracing/StackData.go:43
        D:/Projects/go/src/github.com/Jim3Things/CloudChamber/internal/tracing/client/tracing.go:44
        D:/Projects/go/src/github.com/Jim3Things/CloudChamber/vendor/google.golang.org/grpc/call.go:35
        D:/Projects/go/src/github.com/Jim3Things/CloudChamber/pkg/protos/services/stepper.pb.go:194
        D:/Projects/go/src/github.com/Jim3Things/CloudChamber/internal/clients/timestamp/timestamp.go:87
      @   1: [I] () Operation "logout", user "admin", session &{ map[session-id:1] 0xc00045ad80 false 0xc0002686e0 CC-Session}
        D:/Projects/go/src/github.com/Jim3Things/CloudChamber/internal/tracing/StackData.go:43
        D:/Projects/go/src/github.com/Jim3Things/CloudChamber/internal/tracing/spanEx.go:176
        D:/Projects/go/src/github.com/Jim3Things/CloudChamber/internal/services/frontend/users.go:364
        D:/Projects/go/src/github.com/Jim3Things/CloudChamber/internal/services/frontend/session_manager.go:193
        D:/Projects/go/src/github.com/Jim3Things/CloudChamber/internal/services/frontend/users.go:359

internal/services/frontend/users.go
Jim3Things: It doesn't add a trace entry, but the data here is pretty much debug only so it can be reduced and moved to the default.

internal/tracing/spanEx.go
Jim3Things: I was kind of inconsistent here.  checking for zero len is good practice, but not a case that we ever have, so I didn't follow through.  I'll add the zero len check to the others.

And the single element case no longer is only a string, so the fmt string, a ...interface{} won't work.  It must be a string if it is a multi-element case, which is just let the sprintf handle.

internal/tracing/spanEx.go
Jim3Things: Based on usage, yes, but I'll add the check.

MikeCJohnsonUk: I was actually just asking as I had no idea what the language guaranteed. I could see, and was happy, that 'a' might have no entries, but I wasn't sure what happened if invoked with "formatIf()" : would the compile fail, would it succeed and the called site receive a nil 'a', or a non-nil 'a' which was empty?

internal/tracing/spanEx.go
MikeCJohnsonUk: It was the inconsistency that caught my attention.

I'd actually be ok with a panic on the 0 case if we shouldn't have it. Your call.

internal/services/frontend/users.go

internal/tracing/exporters/common.go
MikeCJohnsonUk: Ok

MikeCJohnsonUk (approved):

PR #162, opened 2020-10-09
Add config support for trace retention limit

Jim3Things:
... also update a few span names in stepper.go


MikeCJohnsonUk (approved):
With a number of the test/verification function taking lots of numeric values, I really found myself wishing for keyword macros to make a resurgence...

PR #167, opened 2020-10-13
Inventory simulation - step 2

Jim3Things:
This contains the initial TOR simulation.  It is currently similar to the PDU, but that is likely to change as we go forward.  Again, like with the PDU change, this only accounts for the repair operations.  The status calls are currently not implemented.

### DIFF
--- a/internal/services/inventory/pdu_test.go
+++ b/internal/services/inventory/pdu_test.go
@@ -52,6 +52,7 @@ func (ts *PduTestSuite) TestCreatePdu() {
 
 	for _, c := range p.cables {
 		assert.False(c.on)
+		assert.False(c.faulted)
 	}
 }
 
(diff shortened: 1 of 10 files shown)


internal/services/inventory/pdu_test.go
```diff
@@ -246,13 +244,166 @@ func (ts *PduTestSuite) TestPowerOnBlade() {
...
+	}
+
+	go func() {
+		r.pdu.Receive(ctx, msg, rsp)
+	}()
+
+	res := common.CompleteWithinInterface(rsp, time.Duration(1) * time.Second)
+	assert.NotNil(res)
+
+	repairResp, ok := res.(*services.InventoryRepairResp)
```
MikeCJohnsonUk, line 289: I've not seen this construct before (at least I don't remember seeing it) so I'm afraid you're going to have to educate me a little. It looks like it's possibly a function call using a field from "res", but if "(*services.InventoryRepairResp)" is the parameters, where's the function field name/address? However, the "*services.InventoryRepairResp" would seem to be a pointer to an InventoryRepairResp struct, so the "res.()" confused me. 

Or is this a simply a case of precedence manipulation in a field reference where I might otherwise have expected "*res.services.InventoryRepairResp" but because res is an interface...?

Jim3Things: It's a testable cast - it tries to cast to the type.  If it does, then the first return value is the cast result and ok is true.  If it doesn't, then ok is false (and I think the first is nil).

MikeCJohnsonUk (approved):

PR #168, opened 2020-10-13
Clean up GRPC tracing client metadata

Jim3Things:
... which includes propagating the infrastructure value across the GRPC call.


internal/tracing/client/tracing.go
```diff
@@ -68,3 +70,18 @@ func decodeGrpcErr(err error) (log.Severity, string) {
 
 	return sev, msg
 }
+
+// parentIsInfra sets returns the string value that denotes whether the current
```
MikeCJohnsonUk, line 74: Comment doesn't quite read right. Missing word "and" perhaps?

Jim3Things: yeah, missed edit.  Fixed.

MikeCJohnsonUk (approved):

PR #170, opened 2020-10-21
Add linked-to support to tracing subsystem

Jim3Things:
This adds tracking of linked spans with enough context to stitch the caller sequence back together.  Linked spans are more loosely coupled than parent/child spans.  Formatting of events and spans now includes link information, but there is no attempt to restructure the io-span tree in the iow forwarder.  The expectation is that the UI log display will stitch these together in real time.

### DIFF
--- a/internal/services/inventory/pdu_test.go
+++ b/internal/services/inventory/pdu_test.go
@@ -6,10 +6,10 @@ import (
 	"time"
 
 	"github.com/stretchr/testify/suite"
-	"go.opentelemetry.io/otel/api/trace"
 
 	"github.com/Jim3Things/CloudChamber/internal/common"
 	"github.com/Jim3Things/CloudChamber/internal/sm"
+	"github.com/Jim3Things/CloudChamber/internal/tracing"
 	"github.com/Jim3Things/CloudChamber/internal/tracing/exporters"
 	ct "github.com/Jim3Things/CloudChamber/pkg/protos/common"
 	pb "github.com/Jim3Things/CloudChamber/pkg/protos/inventory"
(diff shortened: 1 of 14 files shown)


Jim3Things:
I have two options that I'd love opinions on.  First, I could track the active spans (at least the active SpanEx's) for a process and turn a linked-to relationship into a parent/child relationship if the source span is also active in that process.  I've not done that, instead always leaving the linked-to relationship loose.  Should I add this?

Second, while I've not done it yet, I expect I'll want to also add 'continuation links' - ones where the source span stops and expects that the linked span will pick up and continue its work.  E.g. if a state machine is waiting for a timer expiry before continuing.  I'd expect that these would show up in the UI as adjoining peer spans in the tree, rather than tail end children.  Again, does this sound right to you?

internal/services/inventory/pdu_test.go
```diff
@@ -44,6 +58,21 @@ func (ts *PduTestSuite) completeWithin(ch <-chan *sm.Response, delay time.Durati
 	}
 }
 
+func execute(ctx context.Context, msg *sm.Envelope, action func(ctx2 context.Context, envelope *sm.Envelope)) {
```
MikeCJohnsonUk, line 61: Just for my education, does the ctx2 parameter actually need to have a distinct name given that it's just part of the signature for the action parameter, not a parameter to the execute function itself?

internal/services/stepper_actor/adapter.go
```diff
@@ -160,7 +161,13 @@ func msgToError(msg interface{}, err error) (interface{}, error) {
...
 // Get the current span context and encode it into the outgoing message header
 func actorContext(ctx context.Context) *actor.RootContext {
-	spanContext := trace.SpanFromContext(ctx).SpanContext()
+	span := trace.SpanFromContext(ctx)
+	tag, ok := tracing.GetAndMarkLink(span)
+	if ok {
+		tracing.AddLink(ctx, tag)
+	}
+
+	spanContext := span.SpanContext()
```
MikeCJohnsonUk, line 170: Not sure if pointing out this is (most of) the gits of the NewEnvelope() func is worth it. Don't think it's really abstractable, but the itch to mention it was irresistible.

MikeCJohnsonUk (approved):

internal/services/inventory/pdu_test.go
Jim3Things: No, it doesn't have to be different.  I did that just be clear for myself.

internal/services/stepper_actor/adapter.go
Jim3Things: Uhm, I couldn't parse your comment.  ???

PR #174, opened 2020-10-24
Further improvements to test structure and startup

Jim3Things:
Initialize the clients only once, fix tracking of listener and buffer connection in test startup, restructure timestamp client.

Note I'd like feedback on the timestamp client.  If this looks useful I'll convert the sink client as well.

### DIFF
--- a/internal/clients/trace_sink/trace_sink.go
+++ b/internal/clients/trace_sink/trace_sink.go
@@ -24,7 +24,7 @@ type TraceData struct {
 func InitSinkClient(name string, opts ...grpc.DialOption) {
 	dialName = name
 
-	dialOpts = append(dialOpts, opts...)
+	dialOpts = append([]grpc.DialOption{}, opts...)
 }
 
 // Reset forcibly resets the trace sink to its initial state.  This is intended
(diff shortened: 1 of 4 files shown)


internal/clients/trace_sink/trace_sink.go
```diff
@@ -24,7 +24,7 @@ type TraceData struct {
 func InitSinkClient(name string, opts ...grpc.DialOption) {
 	dialName = name
 
-	dialOpts = append(dialOpts, opts...)
+	dialOpts = append([]grpc.DialOption{}, opts...)
```
MikeCJohnsonUk, line 27: Is it worth adding a trap in here to detect the second/subsequent call? 

Question also applies to all the init routines I guess.

MikeCJohnsonUk (approved):
Actually, yes, I do like the timestamp changes. It does provide a clean way to indicate the two fundamental states and there is a simple transition from one to the other.

internal/clients/trace_sink/trace_sink.go
Jim3Things: I think that this gets easier with the timestamp pattern.  I'll see if I can do this better in the next rev.

PR #178, opened 2020-11-08
Add blade boot sequence

Jim3Things:
This adds the blade boot sequence. This is a delayed effect of powering on, which means that it is triggered by a timer. That forced a refactoring of the other unit tests.

This also fixes an issue with linked spans that are started while another span is active in the context.

STill to be done: Tor messages are incomplete, get status is missing. Workload state machine is missing. GRPC service is missing.


internal/services/inventory/blade.go
```diff
@@ -14,109 +16,343 @@ type workload struct {
...
+	// there is one.
+	activeTimerID int64
+
+	// matchTimerExpiry is the blade supplied ID embedded in the timer expired
+	// message.
+	matchTimerExpiry int
 }
 
 const (
+	// bladedOffState is current when the blade has no simulated power.
```
MikeCJohnsonUk, line 91: minor typo - bladeOffState with just the one 'd'

internal/services/inventory/blade.go
```diff
@@ -14,109 +16,343 @@ type workload struct {
...
+	bladeBootDelayMax = int64(5)
+)
+
+// capacity defines the consumable and capability portions of a blade or
+// workload.
+type capacity struct {
+	// consumables are named units of capacity that are used by a workload such
+	// that the amount available to other workloads is reduced by that amount.
+	// For example, a core may only be used by one workload at a time.
+	consumables map[string]float64
```
MikeCJohnsonUk, line 41: why float64 vs int64? I worry about seeing precision errors along the line of we have "3.999999..99" cores etc. If this is purely local/internal, then it may well not be a problem anyway.

internal/services/frontend/TestSuiteCore.go
```diff
@@ -195,60 +183,13 @@ func (ts *testSuiteCore) doLogout(user string, cookies []*http.Cookie) *http.Res
 }
 
 // ensureServicesStarted handles the various components that can only be set or
-// initialized once.  This includes flag parsing and support service startup.
+// initialized once.
 func (ts *testSuiteCore) ensureServicesStarted() {
-	if s == nil {
+	if !initServiceDone {
 		require := ts.Require()
```
MikeCJohnsonUk, line 189: I hate to be pedantic here, but are we sure that this is never going to be called from simultaneous threads (vs. an unknown number of sequential threads)?

internal/services/inventory/blade.go
Jim3Things: fixed

internal/services/frontend/TestSuiteCore.go
Jim3Things: Yes, this will not be done in parallel.  Both gotest and testify suite do serial execution by default.  We, of course, have parallel execution behind the scenes once the tests start, but nothing will cause parallel activity before that.

internal/services/inventory/blade.go
Jim3Things: It's currently local.  I did this to give us a place to look at fractional use - e.g. fractional cores.  I think it is reasonable for the original definition's capacity data to be integers (as there aren't really fractional actual cores, e..g), but I think we'll want the flexibility of floats for the usage.  At least for now.

MikeCJohnsonUk (approved):

PR #179, opened 2020-11-08
Override log severity by method on the GRPC client

Jim3Things:


internal/tracing/client/severityOverride.go
```diff
@@ -0,0 +1,33 @@
...
+// overrideSeverity looks in the table of method names that are designated to
+// have specific severity overrides.  If a match with the supplied method name
+// and with the supplied severity is found, the returned severity is from the
+// override value, otherwise it is the one supplied.
+func overrideSeverity(method string, sev log.Severity) log.Severity {
+	for _, o := range overrides {
+		if o.matchMethod == method && o.matchSeverity == sev {
+			return o.severity
+		}
+	}
```
MikeCJohnsonUk, line 30: I presume this list is expected to be small so an exhaustive search won't take long?

MikeCJohnsonUk (approved):

internal/tracing/client/severityOverride.go
Jim3Things: Yes.

PR #181, opened 2020-11-12
Initial pass at storing inventory definition in the store

MikeCJohnsonUk:
This is still a work-in-progress and is not yet active in production.
This change contains the protobuf file for the definition, create and read routines for zones, racks, pdus, tors, and blades along with some basic tests to exercise the create and read functionality. It also includes templates for subsequent routines for update and delete.

Also included is the planned WithXxxx() option modifiers as templates, but as yet these are not active.

There are also a number of routines which have been added in preparation for future changes but which are not yet active.

The purpose of this PR is publish the intentions for this chunk of feature set to ensure everything is headed in the right direction.

Next steps
- add tests for the UpdateXxxx() routines and ensure everything works as expected
- add tests for the DeleteXxxx() routines and ensure everything works as expected
- add the body for the ListXxxx() routines along with tests to ensure everything works as expected
- add ability to read the definition from a file, and update the store to match the file content.
- add the ability to specify the WithRevision() option for ReadXxx() and ListXxxx() routines.
- add the ability to specify the WithRevision() option for UpdateXxx() and DeleteXxxx() routines.
- cut-over from current inventory definition load to store based inventory definitions.


internal/clients/store/storeapi.go
```diff
@@ -76,8 +175,44 @@ type Request struct {
 	Reason     string
 	Records    map[string]Record
 	Conditions map[string]Condition
+	Actions    map[string]Action
 }
 
+
+/*
```
Jim3Things, line 182: I know that we're using IDEs that colorize, but this seems a bit subtle of a trigger for noticing that this whole section is actually a comment.  Can you make it more visible?

internal/clients/store/storeapi.go
```diff
@@ -203,6 +338,57 @@ func (store *Store) Create(ctx context.Context, r KeyRoot, n string, v string) (
...
+	if err != nil {
+		for k := range request.Records {
+			if err == ErrStoreAlreadyExists(k) {
+				n := getNameFromKeyRootAndKey(r, k)
+				return RevisionInvalid, ErrStoreAlreadyExists(n)
+			}
+		}
+	}
+
+	if err != nil {
```
Jim3Things, line 381: Can't this be folded up into the check at line 372?  I think that if it ever falls out of the loop at 373 it will end up at line 382...

internal/clients/store/storeapi.go
```diff
@@ -203,6 +338,57 @@ func (store *Store) Create(ctx context.Context, r KeyRoot, n string, v string) (
...
+
+	for n, v := range *kvs {
+		k := getKeyFromKeyRootAndName(r, n)
+		request.Records[k] = Record{Revision: RevisionInvalid, Value: v}
+		request.Conditions[k] = ConditionCreate
+	}
+
+	resp, err := store.WriteTxn(ctx, request)
+
+	// Need to strip the namespace prefix and return something described
```
Jim3Things, line 369: Nit: can you specify that this only applies to the loop below?

internal/clients/store/storeapi.go
```diff
@@ -343,6 +529,52 @@ func (store *Store) Read(ctx context.Context, kr KeyRoot, n string) (value *stri
 	}
 }
 
+// Update is a function to conditionally update a value for a single key
+//
+func (store *Store) Update(ctx context.Context, r KeyRoot, n string, rev int64,	v string) (revision int64, err error) {
```
Jim3Things, line 534: odd spacing

internal/config/InventoryReader.go
```diff
@@ -174,3 +173,188 @@ func toExternalZone(xfr *zone) (*pb.ExternalZone, error) {
...
+		if _, ok := err.(viper.ConfigFileNotFoundError); ok {
+			err = fmt.Errorf("no inventory definition found at %s/%s (%s)",
+				path,
+				defaultDefinitionFile,
+				defaultConfigType)
+		} else {
+			err = fmt.Errorf("fatal error reading definition file: %s", err)
+		}
+
+		span.AddEvent(ctx, err.Error())
```
Jim3Things, line 202: Glad to see the conversion to the CC tracing library - the AddEvent calls should also be converting to tracing.Error (or other tracing.Xxx, depending)

internal/config/InventoryReader.go
```diff
@@ -174,3 +173,188 @@ func toExternalZone(xfr *zone) (*pb.ExternalZone, error) {
...
+			Condition: pb.DefinitionZone_operational,
+			Location:  "DC-PNW-0",
+			Notes:     "Base zone",
+		},
+		Racks: make(map[string]*pb.DefinitionRackInternal),
+	}
+
+	// For each rack in the supplied configuration, create rack in the
+	// zone. Each rack has some details, a set of Pdus, a set of Tors, 
+	//and a set of blades.
```
Jim3Things, line 250: nit: spacing.

internal/config/InventoryReader.go
```diff
@@ -174,3 +173,188 @@ func toExternalZone(xfr *zone) (*pb.ExternalZone, error) {
...
+			// For the given blade index, add a matching connection
+			// in the PDU to allow power for the blade to be
+			// connected and controlled.
+			//
+			zone.Racks[r.Name].Pdus[0].Ports[b.Index] = &pb.DefinitionPowerPort{
+				Wired: true,
+				Item:  &pb.DefinitionItem{
+					Type: pb.DefinitionItem_pdu,
+					Id: 0,
+					Port: 0,
```
Jim3Things, line 331: Is this right?  Shouldn't something point to the blade?

internal/config/InventoryReader.go
```diff
@@ -174,3 +173,188 @@ func toExternalZone(xfr *zone) (*pb.ExternalZone, error) {
...
+			// For the given blade index, add a matching connection
+			// in the TOR to allow a network for the blade to be
+			// connected and controlled.
+			//
+			zone.Racks[r.Name].Tors[0].Ports[b.Index] = &pb.DefinitionNetworkPort{
+				Wired: true,
+				Item:  &pb.DefinitionItem{
+					Type: pb.DefinitionItem_tor,
+					Id: 0,
+					Port: 0,
```
Jim3Things, line 344: ditto

internal/config/InventoryReader_test.go
```diff
@@ -107,9 +110,125 @@ func TestInventoryValidateBlade(t *testing.T) {
...
-	_, err := ReadInventoryDefinition("./testdata/BadYamlValidate")
+	_, err := ReadInventoryDefinition(context.Background(), "./testdata/BadYamlValidate")
 	require.NotNil(t, err)
 	assert.Equal(t,
 		"In rack \"rack1\": the field \"Blades[2].Cores\" must be greater than or equal to 1.  It is 0, which is invalid",
 		err.Error())
 }
+
+
+
```
Jim3Things, line 121: fewer blank lines

internal/config/InventoryReader_test.go
```diff
@@ -107,9 +110,125 @@ func TestInventoryValidateBlade(t *testing.T) {
...
+		assert.True(t, r.Details.Enabled)
+		assert.Equal(t, inventory.Definition_operational, r.Details.Condition)
+		assert.Equal(t, "DC-PNW-0-" + name, r.Details.Location)
+		assert.Equal(t, "RackName: " + name, r.Details.Notes)
+
+		assert.Equal(t, 1, len(r.Pdus))
+		assert.Equal(t, 1, len(r.Tors))
+		assert.Equal(t, 2, len(r.Blades))
+
+
```
Jim3Things, line 162: Test the cables somewhere in here.

internal/services/frontend/DBInventory.go
```diff
@@ -54,42 +91,176 @@ var dbInventory *DBInventory
 // connected to the store in order to persist the inventory read from an external
 // definition file
 //
-func InitDBInventory(cfg *config.GlobalConfig) error {
+func InitDBInventory(ctx context.Context, cfg *config.GlobalConfig) (err error) {
```
Jim3Things, line 94: Would it be easier to test, and if so, is this a good time, to have Init return a pointer to the struct, rather than assume the global?

internal/services/frontend/DBInventory.go
```diff
@@ -54,42 +91,176 @@ var dbInventory *DBInventory
...
 // function with each entry.
-func (m *DBInventory) Scan(action func(entry string) error) error {
-	m.Mutex.Lock()
-	defer m.Mutex.Unlock()
+func (m *DBInventory) ScanRacks(action func(entry string) error) error {
+	m.mutex.RLock()
+	defer m.mutex.RUnlock()
 
-	for name := range dbInventory.Zone.Racks {
+	for name := range m.Zone.Racks {
```
Jim3Things, line 263: good catch!

pkg/protos/inventory/external.proto
```diff
@@ -6,7 +6,7 @@
 
 syntax = "proto3";
 
-import "github.com/Jim3Things/CloudChamber/pkg/protos/common/capacity.proto";
+import "pkg/protos/common/capacity.proto";
```
Jim3Things, line 9: Why this change?

pkg/protos/inventory/definition.proto
```diff
@@ -0,0 +1,289 @@
...
+
+        // The set of racks forming an individual zone. All racks in the zone are affected
+        // by the properties of the zone or any conditions affecting the zone. Each rack is
+        // defined by a string name within the zone, and that name is used here as a key.
+        //
+        map<string, rack_internal> racks = 10;
+    }
+}
+
+message zonemap {
```
Jim3Things, line 285: Nit, but might be better as either 'zones' or 'zone_map'

pkg/protos/inventory/definition.Validate.go
```diff
@@ -0,0 +1,284 @@
...
+	return nil
+}
+
+// Verify is a method that verifies that the associated DefinitionTor instance
+// is semantically legal
+//
+// For exameple, check for errors such as the tor being wired to itself, the
+// port being wired but without an associated item etc.
+//
+func(x *DefinitionTor) Verify() error {
```
Jim3Things, line 123: Who calls Verify, and is there a reason why it is not part of the normal Validate processing?

pkg/protos/inventory/definition.proto
```diff
@@ -0,0 +1,289 @@
+// This file contains the definitions used by configuration definitions for describing the simulated
+// inventory.  These definitions describe the items, but do not define any observed or actual status for them.
+//
+// Note that we may, at some point, want to add an initial status definition in order to define inventory that begins
+// in a partially failed manner.
+
+syntax = "proto3";
+
+import "pkg/protos/common/capacity.proto";
```
Jim3Things, line 9: We use the full path elsewhere, I'm not sure why the short path here?

Jim3Things (changes requested):

internal/clients/store/storeapi.go
MikeCJohnsonUk: Done

internal/clients/store/storeapi.go
MikeCJohnsonUk: Yup. Done.

internal/clients/store/storeapi.go
MikeCJohnsonUk: Makes sense. Done.

internal/clients/store/storeapi.go
MikeCJohnsonUk: Ha! An embedded tab which with all my editors hid itself. Good spot.

Fixed,

internal/config/InventoryReader.go
MikeCJohnsonUk: I've fixed this and converted the others. I believe I have the usage correct, but would appreciate the confirmation.

Done.

internal/services/frontend/DBInventory.go
MikeCJohnsonUk: Yeah, I was pleased with myself on this one.

internal/config/InventoryReader.go
MikeCJohnsonUk: Done.

internal/config/InventoryReader.go
MikeCJohnsonUk: Correct. What the 'Item' here should be is something along the lines of

				Item:  &pb.DefinitionItem{
					Type: pb.DefinitionItem_blade,
					Id: b.Index,
					Port: 0,
					},
which says that this pdu port we are adding is connected to a blade with index b.Index and port 0 (for now) on that blade. There is an analogous construct for the tor.

internal/config/InventoryReader_test.go
MikeCJohnsonUk: Done.

pkg/protos/inventory/external.proto
MikeCJohnsonUk: Took me a while to remember.

In the new definition.proto file, if I use the full path, the CLI driven compilation is fine, but VsCode objects claiming the file "was not found or had errors".

In this location, with the same import path, everything looks fine and I removed the prefix to see what impact it had, the answer being none. I just neglected to revert this change.

Not sure what to do with the import path in definition.proto though. I'm open to suggestions. though.

MikeCJohnsonUk: Update: it seems VsCode also complains here with the prefixed import path.

I've done some digging around, and I *think* the import path should be a relative directory path rather than a package name. That is, I believe we should be using 

import "pkg/protos/common/capacity.proto";

and not, as we currently have

import "github.com/Jim3Things/CloudChamber/pkg/protos/common/capacity.proto";

The "option go_package= xxx" should remain unchanged.


I suspect you may have been using an IDEA option to locate the proto files somewhere?

pkg/protos/inventory/definition.proto
MikeCJohnsonUk: See discussion above, but in summary, I believe this is a file path reference, and not a package location and so should be relative to the base of the project directory.

pkg/protos/inventory/definition.proto
MikeCJohnsonUk: changed to "region" as it is somewhat more relevant!

pkg/protos/inventory/definition.Validate.go
MikeCJohnsonUk: I was calling it under slightly different circumstance which didn't have the same requirements as validate.

I've changed this to be a common internal routine which does all the validation and then a couple of outer functions dealing the the parameters differences and a simple specialization in one case. This means both parts get the complete generalized check.

I think this fixes the problem.

internal/config/InventoryReader.go
```diff
@@ -174,3 +169,184 @@ func toExternalZone(xfr *zone) (*pb.ExternalZone, error) {
...
+	region := &pb.Region{Zones: make(map[string]*pb.DefinitionZoneInternal)}
+
+	// Since we only have a single zone at present, there is no loop
+	// here. But there will be eventually.
+	//
+	zone := &pb.DefinitionZoneInternal{
+		Details: &pb.DefinitionZone{
+			Enabled:   true,
+			Condition: pb.DefinitionZone_operational,
+			Location:  "DC-PNW-0",
```
Jim3Things, line 234: Fyi: this would be best to be a string constant, as I expect the location is going to ultimately either come from inventory def yaml or from config.

internal/services/frontend/DBInventory.go
```diff
@@ -170,10 +342,839 @@ func (m *DBInventory) buildSummary() {
...
+//
+func (m *DBInventory) ListPdus(ctx context.Context, zone string, rack string, options ...InventoryOption) (map[string]*DefinitionPdu, int64, error) {return nil, InvalidRev, nil}
+
+// ListTors returns the basic records for the TORs in the specified rack.
+//
+func (m *DBInventory) ListTors(ctx context.Context, zone string, rack string, options ...InventoryOption) (map[string]*DefinitionTor, int64, error) {return nil, InvalidRev,  nil}
+
+// ListBlades returns the basic records for the blades in the specified rack.
+//
+func (m *DBInventory) ListBlades(ctx context.Context, zone string, rack string, options ...InventoryOption) (map[string]*DefinitionBlade, int64, error) {return nil, InvalidRev, nil}
```
Jim3Things, line 612: Nit: formatting - lines too long...

internal/services/frontend/DBInventory_test.go
```diff
@@ -0,0 +1,572 @@
...
+	assert.Equal(blade.Capacity.MemoryInMb,             b.Capacity.MemoryInMb)
+	assert.Equal(blade.Capacity.DiskInGb,               b.Capacity.DiskInGb)
+	assert.Equal(blade.Capacity.NetworkBandwidthInMbps, b.Capacity.NetworkBandwidthInMbps)
+	assert.Equal(blade.Capacity.Arch,                   b.Capacity.Arch)
+	assert.Equal(blade.Capacity.Accelerators,           b.Capacity.Accelerators)
+}
+
+
+// func (ts *DBInventoryTestSuite) TestLoadFromStore() {
+// 	assert := ts.Assert()
```
Jim3Things, line 489: Presumably these commented out lines will be deleted at some point?

internal/services/frontend/frontend.go
```diff
@@ -29,6 +29,7 @@ import (
 	"google.golang.org/grpc"
 
 	"github.com/Jim3Things/CloudChamber/internal/clients/store"
+	clients "github.com/Jim3Things/CloudChamber/internal/clients/timestamp"
 	ts "github.com/Jim3Things/CloudChamber/internal/clients/timestamp"
```
Jim3Things, line 33: aren't these duplicates?

pkg/protos/common/errors.go
```diff
@@ -107,3 +159,55 @@ func (e ErrMinLenString) Error() string {
...
+}
+
+func (e ErrItemMissingValue) Error() string {
+	return fmt.Sprintf("the field %q for %q port %q must have a value.",
+		e.Field,
+		e.Item,
+		e.Port)
+}
+
+// ErrInvalidItemSelf signals that the specified item has wired a port to itsel.
```
Jim3Things, line 196: itself.

pkg/protos/inventory/definition.Validate.go
Jim3Things: ok

pkg/protos/inventory/external.proto
Jim3Things: we followed up offline - the issue I had previously was due to the pre-makefile build file.  If this works then we should go the relative version throughout (in another set of changes).

internal/config/InventoryReader_test.go
MikeCJohnsonUk: Done

internal/services/frontend/DBInventory.go
MikeCJohnsonUk: In principle, yes I would agree. If there are no objection, I think I'd prefer to leave this to the next phase of these related changes.

pkg/protos/inventory/external.proto
MikeCJohnsonUk: Seems to be working so far. If after merging into master, they continue to work for everyone else, I'll fix this up globally.

internal/config/InventoryReader.go
MikeCJohnsonUk: This was just an arbitrary string I added to make sure it has a recognizable value whilst debugging. I fully expect this will be sourced from the inventory defn yaml file at some point in the future.

Is this something that needs a change at this time?

internal/services/frontend/DBInventory.go
MikeCJohnsonUk: Shortened, but while better, I'm still not really happy about how it ended up.

internal/services/frontend/DBInventory_test.go
MikeCJohnsonUk: The approach I was planning has all changed, so for now I've removed these and something will take their place when the replacement loading logic is added.

Done.

internal/services/frontend/frontend.go
MikeCJohnsonUk: Huh. I suspect these are remnants of a troubled merge that I failed to spot.

Fixed. (by removing the version on line 32.)

pkg/protos/common/errors.go
MikeCJohnsonUk: Done.

internal/config/InventoryReader.go
Jim3Things: No, this can wait.

Jim3Things (approved):

internal/config/InventoryReader.go
Jim3Things: Seems to be

PR #184, opened 2020-11-23
Clean up trace and timestamp client initialization

Jim3Things:
This is a follow-on from an earlier partial change to handle proper initialization code for these two clients.

They now return an 'already initialized' error that is used in production to indicate an internal error.  Test suites expect to get this, and that it is harmless, so the error is ignored in those cases.

They both cache the grpc connection and clients, replacing them after a failure.  There is a unit test for the timestamp client that verifies this handling.

There is a common pattern and almost perfectly common code between these initializations.  I may factor that out in the future, even though it is a small amount of code.

### DIFF
--- a/internal/services/frontend/frontend.go
+++ b/internal/services/frontend/frontend.go
@@ -136,17 +136,21 @@ func initHandlers() error {
 // initClients sets up the internal service clients used by the frontend
 // handlers themselves.
 func initClients(cfg *config.GlobalConfig) error {
-	ts.InitTimestamp(
+	err := ts.InitTimestamp(
 		cfg.SimSupport.EP.String(),
 		grpc.WithInsecure(),
 		grpc.WithUnaryInterceptor(ct.Interceptor))
 
-	tsc.InitSinkClient(
+	if err != nil {
+		return err
+	}
+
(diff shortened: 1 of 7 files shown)


internal/services/frontend/frontend.go
```diff
@@ -136,17 +136,21 @@ func initHandlers() error {
...
-	ts.InitTimestamp(
+	err := ts.InitTimestamp(
 		cfg.SimSupport.EP.String(),
 		grpc.WithInsecure(),
 		grpc.WithUnaryInterceptor(ct.Interceptor))
 
-	tsc.InitSinkClient(
+	if err != nil {
+		return err
+	}
```
MikeCJohnsonUk, line 146: Separate standalone test vs single statement, inline test for err?

Only highlighting difference in construction vs (say) lines 27-32 in inventory.go.

No particular preference for either style.

MikeCJohnsonUk (approved):

internal/services/frontend/frontend.go
Jim3Things: I've been torn on the cases where there are a lot of lines in the call.  Not clear to me which pattern is best in those cases.

PR #185, opened 2020-11-29
Clean up test sync issues

Jim3Things:
The state machine mechanisms are correctly synchronized with respect to themselves, but the actual state change is placed after the response is issued.  That means that the unit tests, which reach around the normal synchronization, need to be aware that it may take a brief time for the state change to show up.  This was missed outside of the blade unit tests.  This is fixed in this change.

Likewise, the tests now use the constants for the state IDs, rather than in line strings.

### DIFF
--- a/internal/services/inventory/pdu_test.go
+++ b/internal/services/inventory/pdu_test.go
@@ -24,14 +24,14 @@ func (ts *PduTestSuite) TestCreatePdu() {
 
 	r := newRack(context.Background(), ts.rackName(), rackDef, ts.timers)
 	require.NotNil(r)
-	assert.Equal("awaiting-start", r.sm.CurrentIndex)
+	assert.Equal(rackAwaitingStartState, r.sm.CurrentIndex)
 
 	p := r.pdu
 	require.NotNil(p)
 
 	assert.Equal(2, len(p.cables))
 
-	assert.Equal("working", p.sm.CurrentIndex)
+	assert.Equal(pduWorkingState, p.sm.CurrentIndex)
 
 	for _, c := range p.cables {
 		assert.False(c.on)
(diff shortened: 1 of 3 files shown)


internal/services/inventory/pdu_test.go
```diff
@@ -132,14 +145,20 @@ func (ts *PduTestSuite) TestPowerOffPduTooLate() {
 		assert.True(c.on)
 	}
 
-	assert.Equal("working", r.pdu.sm.CurrentIndex)
+	// Verify that it did not change - should never need to wait for this.
+	assert.Equal(pduWorkingState, r.pdu.sm.CurrentIndex)
```
MikeCJohnsonUk, line 149: There may well be no need to wait at present, but is that always going to be the case?

internal/services/inventory/tor_test.go
```diff
@@ -121,7 +121,17 @@ func (ts *TorTestSuite) TestConnectBlade() {
...
+		return r.tor.sm.CurrentIndex == torWorkingState
+	})
+
+	require.True(ok, "state is %v", r.tor.sm.CurrentIndex)
+
+	ok = ts.waitForStateChange(func() bool {
+		return r.blades[0].sm.CurrentIndex == bladeOffDiscon
+	})
+
+	require.True(ok, "state is %v", r.blades[0].sm.CurrentIndex)
```
MikeCJohnsonUk, line 134: Given we're talking about different objects here, would it be possible to distinguish them with indicating which object each require.Xxx() call is verifying? Just as a convenience. Alternatively, would it be worth including the object type in the tags, e.g. torWorkingState -- torWorking and so on?

internal/services/inventory/pdu_test.go
Jim3Things: Yes, it should always be the case.  We're checking that the state did not change, so there must be no transition to wait for.

internal/services/inventory/tor_test.go
Jim3Things: I could do, but I thought that the line number would be sufficient... ?

MikeCJohnsonUk: I don't think I'd bother then, after all, it's not a trace and we only expect to see one of these at most.

internal/services/inventory/pdu_test.go
MikeCJohnsonUk: Ok

MikeCJohnsonUk (approved):

PR #188, opened 2020-12-09
Convert stepper_test to use testify suite

Jim3Things:


internal/services/stepper/stepper_test.go
```diff
@@ -26,342 +24,434 @@ import (
...
 
-	_, err = client.SetPolicy(ctx, &pb.PolicyRequest{Policy: pb.StepperPolicy_Manual, MeasuredDelay: &duration.Duration{Seconds: 2}, MatchEpoch: -1})
-	assert.NotNilf(t, err, "SetPolicy unexpectedly succeeded with an invalid duration: %v", err)
+	_, err := ts.client.SetPolicy(
+		ctx,
+		&pb.PolicyRequest{
+			Policy:        pb.StepperPolicy_Manual,
+			MeasuredDelay: &duration.Duration{Seconds: 1},
+			MatchEpoch:    -1})
+	assert.Error(err)
```
MikeCJohnsonUk, line 189: I'm ok on losing the strings in the assert message, but the problem is that now there's nothing to say why the test is expected to return an error.

Should we move something derived from the string to a comment, either on the call or on the function? Or perhaps a reference in the production code where the allowable combinations are called out? Can we make cross-module doc references? Not sure what would be best here.

MikeCJohnsonUk (approved):

internal/services/stepper/stepper_test.go
Jim3Things: I can see a comment in the test describing what is being tested.

I dumped the text as it was not adding value to most of the asserts and only adding complexity to the writing of the tests.  But yeah, that loses the why when it isn't obvious.

I'm not sure that I'll catch everything, but let me take a stab at adding test case level comments describing what is being tested - at least where it is a testing edge conditions like this.

PR #189, opened 2020-12-13
Create decode.go

manpreet85:
Hi Mike and Jim,

I have attempted to resolve all three tasks. 

I am struggling to clearly think (and write) how to capture "match tests" and then to return. Also how to capture error. Please advise.


internal/services/frontend/decode.go
```diff
@@ -0,0 +1,44 @@
...
+
+	var test []string = strings.Split(source, ",")
+	
+	// Third subtask is to loop through all the test slice elements.
+	
+	for i := 0; i < 3; i++ {
+
+	// Second subtask is to figure out how to test one element in the test slice
+	// -- how to tell if test[0] matches any of the keywords elements, and which one.
+		if (strings.Contains(test[i], keywords[i])) {
```
MikeCJohnsonUk, line 30: Consider what might happen if the source string contains "Undefined, Observed-by-me"?

internal/services/frontend/decode.go
```diff
@@ -0,0 +1,44 @@
...
+
+func decode(keywords []string, source string) ([]bool, error){
+	// First subtask is to get from source to test (convert the source string into the test string slice)
+	// hint: strings package probably has something like split and trim.
+
+	var test []string = strings.Split(source, ",")
+	
+	// Third subtask is to loop through all the test slice elements.
+	
+	for i := 0; i < 3; i++ {
```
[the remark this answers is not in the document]
Jim3Things: I agree with Mike's comments, and want to add something.

If you look at the decode function definition on line 18, it doesn't require a specific length of the keywords array.  Likewise, line 22 doesn't do so either for the test array.  Adding a required length after that is something that you should always look at with suspicion.

And I'd echo Mike's comment about range loops.

internal/services/frontend/decode.go
Jim3Things: Agreed.  Also, this only checks a diagonal cut across the arrays: i.e. test[0] == keywords[0], test[1] == keywords[1], ...

internal/services/frontend/decode.go
```diff
@@ -0,0 +1,44 @@
...
+
+	// Second subtask is to figure out how to test one element in the test slice
+	// -- how to tell if test[0] matches any of the keywords elements, and which one.
+		if (strings.Contains(test[i], keywords[i])) {
+
+			 response = append(response, keywords[i])
```
Jim3Things, line 32: Remember that the result should be an array of bools, where the entries are true if the equivalent entry in keywords was found and false if it wasn't.  This looks to produce a string array containing only the keywords that were matched.

internal/services/frontend/decode.go
```diff
@@ -0,0 +1,44 @@
...
+
+// Input and output for a faiure case
+var badView = "AbaddefinedobservedValue" 
+var badexpected []bool = nil
+
+func decode(keywords []string, source string) ([]bool, error){
+	// First subtask is to get from source to test (convert the source string into the test string slice)
+	// hint: strings package probably has something like split and trim.
+
+	var test []string = strings.Split(source, ",")
```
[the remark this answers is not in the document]
Jim3Things: As a suggestion, you could write a couple of unit tests that try out these test cases, and use the debugger to step through the unit test to see what you have in the test array after the split.

Jim3Things:
You asked about handling errors.  My first thought is that there are two cases that would clearly be errors, and  a third that might be an error.

The first case is where the keywords array has no entries or is nil.  Since that can never succeed, returning an error about the keywords array itself is good.

The second case is that you've have a test array entry where you found no matches in the keywords array.  In that case you would definitely want to return an error, as you have a source string that contains an invalid item.

The last case is more questionable.  Should you return an error if the keywords array contains an entry more than once?  You could either check that at the start of decode and return an error if there are duplicates, or you can not do that and instead handle that in the test loop (by always checking all entries in the keywords array for a match against each test array entry).

And if you do find an error, simply return immediately with nil, <your error>.

Does that help?  Can you think of other error cases?

internal/services/frontend/decode.go
```diff
@@ -0,0 +1,58 @@
...
+	
+	// Third subtask is to loop through all the test slice elements.
+	var outputkeywords []bool := make([]bool, len(keywords))
```
Jim3Things, line 25: You need either var or :=, but not both.

It's good that you're posting the changes at the end of the day.  Keep that up.  A suggestion is to either keep on top of problems as they happen, or to plan to clean them up every little bit of time.  It just helps you see what changes you've been making.

internal/services/frontend/decode.go
```diff
@@ -0,0 +1,58 @@
...
+	// I need to create putput array of right size with all elements set to False
+	
+	for _, t := range test{
+		
+		for i, k := range keywords{
+		
+	// Second subtask is to figure out how to test one element in the test slice
+	// -- how to tell if test[0] matches any of the keywords elements, and which one.
+			if (k == t {
```
Jim3Things, line 34: you don't need the '('.

internal/services/frontend/decode.go
```diff
@@ -0,0 +1,58 @@
...
+			//Set the output element of [i] to true 
+				outputkeywords[i] = true
+			
+				response = append(response, outputkeywords[i])
```
Jim3Things, line 38: do you still need this?  What is your thought behind the response variable?

internal/services/frontend/decode.go
```diff
@@ -0,0 +1,58 @@
...
+
+			else 
+
+			return error
```
Jim3Things, line 42: Can you talk me through this case, and when it happens?

internal/services/frontend/decode.go
manpreet85: response is the array prepared by decode function  that matches and contains the allowed elements based on the array "views"

internal/services/frontend/decode.go
manpreet85: Only case that I thought of was when no element from "expected" array matched the "view" array

internal/services/frontend/decode.go
Jim3Things: yes, it does, but won't that be the same as the test array at the end?

internal/services/frontend/decode.go
Jim3Things: Yes, it it would be an error if an element in the test array fails to match against any element in the keywords array.  But with more than one keyword, every test array entry will fail to match against at least one element in the keywords array.  This test would, if I read it correctly, trigger in those cases as well.

I think you should put together at least one UT and try walking through it.  I think that would help.

internal/services/frontend/decode.go
```diff
@@ -0,0 +1,67 @@
...
+
+	// keywords and source are passed into decode and are immutable.
+	// 1st subtask creates this: test is created from the data in source
+	// 2nd subtack creates this: matches []bool that is created inside decode and 
+	// 2nd subtask starts this: ...filled in by the match tests, and then returned
+	// 3rd subtask completes this: ...filled in by the match tests, and then returned
+			if h != t4{
+				outputkeywords[j] = false
+			}
+			return badexpected, error ()
```
Jim3Things, line 63: Did this work when you ran the two unit tests below?  I assume that TestDecode would have failed, but wanted to check.

This looks like it will always return a failure -- there's nothing to prevent line 63 from executing.  And the outputkeywords array should only be part of the success test - it doesn't need to be modified in this failure test.

The thing you're trying to verify is that if an element in the test array does not match against any element in the keywords array, you should return an error.  The thing is, though, that you can only test one element at a time, so you have to reverse the thinking - namely that if you find an element in the keywords array that matches the element from the test array then you can't fail on that test element.

So, think about putting a new declaration between lines 49 & 50 -- specifically, 'found := false'.  Now think about when found should be set to true, and when you should check what the state of found is in order to decide if you need to return an error.

Jim3Things: [Also, don't return badexpected.  Return nil.  The declarations from lines 9 to 16 were to document what the two unit tests should be looking for]

internal/services/frontend/decode.go
```diff
@@ -0,0 +1,73 @@
...
+	// 2nd subtack creates this: matches []bool that is created inside decode and 
+	// 2nd subtask starts this: ...filled in by the match tests, and then returned
+	// 3rd subtask completes this: ...filled in by the match tests, and then returned
+			if h == t4{
+				found = true
+			}
+		}
+		if found == false {
+			return nil, errors.New("Bad view found")
+		}
```
Jim3Things, line 69: Great!  This should be a good failure check.  I'm assuming that the unit tests run?

This should be the last of the feature functionality that we need for this function.  What is left is the clean up and edge case handling to finish it out as a function that we can use.

For instance, 

- Can you see how to merge this failure check into the success loop above, so you only have one loop through test which itself only has one inner loop through keywords?

- There are a number of preamble vars in lines 8 - 17.  Where should they be?

- There are probably a couple of other unit tests that you can see.  Go ahead and add them.

Anything else?

See what you can get through in your next pass.

Thanks much, Jim.

manpreet85: Yes the unit test run successfully.

internal/services/frontend/decode.go
```diff
@@ -0,0 +1,35 @@
+package frontend
+
+import (
+	"errors"
+	"strings"
+)
+
+var keywords []string = []string  {"Defined", "Actual", "Target", "Observed"}
```
Jim3Things, line 8: I'd move this to the test code as well.  The decode function does, and should, not care what the keywords are.

internal/services/frontend/decode.go
```diff
@@ -0,0 +1,35 @@
...
+
+func decode(keywords []string, source string) ([]bool, error){
```
Jim3Things, line 10: You should add a comment here.

internal/services/frontend/decode.go
```diff
@@ -0,0 +1,35 @@
...
+	
+	var test []string = strings.Split(source, ",")
+	
+	outputkeywords := make([]bool, len(keywords))
```
Jim3Things, line 14: The overall logic is fine.  Two things about this body: first, outputkeywords is a compound word, so should be outputKeywords, or some other descriptive name.  Second, the return on line 33 should be indented correctly.

internal/services/frontend/decode_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+// Input and output for a success case
+var views = "Defined, Observed"
+var response []string
+var expected []bool = []bool {true, false, false, true}
+
+// Input and output for a faiure case
+var badView = "AbaddefinedobservedValue" 
+var multipleview = "Defined, Defined, Observed, Actual"
+var badexpected []bool = nil
```
Jim3Things, line 18: Now that these have moved to the test file, should these be local to the functions that need them?  In general, keeping variables declared with as small a scope as they need is good, so if a variable can move from a module level to a function level that would be good.

manpreet85: More explanation needed, may be during a Skype call.

internal/services/frontend/decode.go
Jim3Things: I'm going to ask that you do a bit more here.  I expect that once it works, we'll be moving the decode function into a more common package and use it in a number of places in CloudChamber.  That means that decode will become Decode, and that the comment describing what it does becomes more important.  Describe what it does, what the expectations on the arguments are, and how the arguments relate to what it does.

internal/services/frontend/decode.go
```diff
@@ -0,0 +1,36 @@
...
+			}
+
+		}
+		
+		if found == false {
+			return nil, errInMatch
+		}
+	}
+
+return outputkeywords, nil
```
Jim3Things, line 35: indent this.

internal/services/frontend/decode_test.go
```diff
@@ -0,0 +1,74 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// Input and output for a success case
+var keywords []string = []string  {"Defined", "Actual", "Target", "Observed"}
+var response []string
```
Jim3Things, line 12: Is this used?  If it is, is it written into?  If it isn't used, please delete it.  If it is, and it is written into, then it should be local to the test that writes into it.

internal/services/frontend/decode_test.go
```diff
@@ -0,0 +1,74 @@
...
+
+	var badView = "AbaddefinedobservedValue" 
+	var badexpected []bool = nil
+
+	got, err := decode(keywords, badView)
+	require.Error(t, err)
+	assert.Equal(t, badexpected, got)
+}
+
+func TestDecodeMultipleView(t *testing.T) {
```
Jim3Things, line 34: Did you run this one of the tests that follow?  I'd expect this one to fail based on my reading, and am looking to see if I've read it wrong.

internal/services/frontend/decode_test.go
```diff
@@ -0,0 +1,74 @@
...
+ // COMPLETELY BLANK SOURCE	"" WHAT HAPPENS then
+func TestDecodeBlankSource(t *testing.T){
+	var blanksource = ""
+	var expected []bool = []bool {false, false, false, false}
+
+	got, err := decode(keywords, blanksource)
+	require.Error(t, err)
+	assert.Equal(t, expected, got)
+}
+ // what happens when the source is NIL
```
Jim3Things, line 51: Why is this one commented out?

internal/services/frontend/decode_test.go
```diff
@@ -0,0 +1,74 @@
...
+// 	var expected []bool = []bool {false, false, false, false}
+
+// 	got, err := decode(keywords, Nilsource)
+// 	require.Error(t, err)
+// 	assert.Equal(t, expected, got)
+//  }
+
+ //What happens when the source or keyword is empty array
+
+ func TestDecodeemptyarraySource(t *testing.T){
```
Jim3Things, line 63: Does this one compile?  the second decode argument is a string, not a string array, so I'm surprised if it compiles.  It was an empty keyword array that you need to test for.

internal/services/frontend/decode_test.go
```diff
@@ -0,0 +1,74 @@
...
+	var emptyArraysource = make ([]string,0, 4) 
+	var expected []bool = []bool {true, false, true, true}
+
+	got, err := decode(keywords, emptyArraysource)
+	require.Error(t, err)
+	assert.Equal(t, expected, got)
+ }
+ //What happens when the source or keyword is nil array
```
Jim3Things, line 71: Just to be clear, it is if source or keyword are set to nil.  I don't think you can set the second argument to nil and compile, in which case that answers that and you don't need a test for it.  But you can set the first argument to nil, and that should be tested.

internal/services/frontend/decode.go
```diff
@@ -0,0 +1,46 @@
...
+	return fmt.Sprintf(
+		"The source %q was not found in keywords %v.",
+		e.source,
+		e.keywords)
+}
+
+// Decode is a function that finds all occurences of the source string in the keyword array.
+// - It returns an error if no occurence of the source string is found in the keyword array.
+// - When there is no error, the true value returned in boolean array represents that the
+// source string is not null and is equal to one or more elements of the keyword array.
```
Jim3Things, line 23: The comment is a lot better, but I think this part can be made a bit clearer.  How about:

- When there is no error, the boolean array describes which keywords were matched.  An element in that array is true if the element with the same index in the keywords array was matched.

Oh, and I'd call it 'keywords' array rather than 'keyword' array, so that the name used in the comment matches the parameter name.

internal/services/frontend/decode.go
```diff
@@ -0,0 +1,46 @@
...
+		t2 := strings.TrimSpace(t)
+		found := false
+		
+		for i, k := range keywords{
+			if k == t2 {
+				outputkeywords[i] = true
+				found = true	
+			}
+		}
+		if found == false {
```
Jim3Things, line 38: add blank line

internal/services/frontend/decode.go
```diff
@@ -0,0 +1,46 @@
...
+		e.keywords)
+}
+
+// Decode is a function that finds all occurences of the source string in the keyword array.
+// - It returns an error if no occurence of the source string is found in the keyword array.
+// - When there is no error, the true value returned in boolean array represents that the
+// source string is not null and is equal to one or more elements of the keyword array.
+func Decode(keywords []string, source string) ([]bool, error){
+	var test []string = strings.Split(source, ",")
+		outputkeywords := make([]bool, len(keywords))
```
Jim3Things, line 26: a nit that got lost in the detail work to date, but outputkeywords is a compound name, so it should be camel cased appropriately - e.g. outputKeywords.  Same for some of the names in the tests.

Jim3Things (commented):
A few small nits, but very close.

Jim3Things (approved):

Jim3Things (approved):

PR #190, opened 2020-12-15
Add getstatus processing to inventory element state machines

Jim3Things:
This change adds handling of status requests to the underlying inventory elements.

It also removes droppedmessage responses in favor of just closing the reply channel.  This is part of now always closing the reply channel after issuing the response to the incoming request.


internal/services/inventory/pdu.go
```diff
@@ -277,12 +300,11 @@ func setPowerForBlade(
 				"after other changed occurred.  The blade's power state remains unchanged.",
 			msg.Target.Describe())
 
-		msg.Ch() <- messages.DroppedResponse(occursAt)
 		break
```
MikeCJohnsonUk, line 303: I know that in general DroppedReponses go away, but in this case should we be telling the caller they blew it?

internal/services/inventory/tor.go
```diff
@@ -188,14 +270,12 @@ func workingSetConnection(ctx context.Context, machine *sm.SM, m sm.Envelope) bo
 				"after other changed occurred.  The blade's network connection "+
 				"state remains unchanged.",
 			msg.Target.Describe())
-
-		msg.Ch() <- messages.DroppedResponse(occursAt)
 		break
```
MikeCJohnsonUk, line 273: As above.

internal/services/inventory/blade_test.go
```diff
@@ -49,6 +49,83 @@ func (ts *BladeTestSuite) issueSetConnection(ctx context.Context, r *Rack, id in
...
+	require.NotNil(p)
+	require.NoError(p.Err)
+
+	ctx, ok := ts.advanceToStateChange(ctx, 5, func() bool {
+		return r.blades[0].sm.CurrentIndex == bladePoweredDiscon
+	})
+
+	require.True(ok, "state is %v", r.blades[0].sm.CurrentIndex)
+
+	// Powered on, but disconnected, so this should fail
```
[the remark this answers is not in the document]
Jim3Things: Sure, I've updated the comment.

I've also updated the ts.completeWithin to indicate if the channel was closed or if it timed out.  While there is no feature difference, the channel closure is to keep the simulation moving along in real time.  So, adding that check was good.

internal/services/inventory/pdu.go
Jim3Things: I can go either way.  I was trying to figure out the real world behavior I was simulating with a message race, but returning an error is fine.  I've done that now.

internal/services/inventory/tor.go
Jim3Things: as above

MikeCJohnsonUk (approved):

PR #195, opened 2020-12-29
Move frontend errors to common errors.go

Jim3Things:
http errors are kept in frontend, as they are specific to HTTP serving.


pkg/errors/formatters.go
```diff
@@ -0,0 +1,20 @@
...
+	return fmt.Sprintf("blade %d in zone %q, rack %q", bladeID, zone, rack)
+}
+
+func pduAddress(zone string, rack string, pdu int64) string {
+	return fmt.Sprintf("pdu %d in zone %q, rack %q", pdu, zone, rack)
+}
+
+func torAddress(zone string, rack string, tor int64) string {
+	return fmt.Sprintf("tor %d in zone %q, rack %q", tor, zone, rack)
+}
```
MikeCJohnsonUk, line 20: Now, *this* is a good idea. I like it! Obvious in hindsight of course...

MikeCJohnsonUk (approved):

PR #203, opened 2021-01-18
Fix how ETag values are constructed and parsed

Jim3Things:
ETag values are quoted strings, so we need to add the extra quotes when sending, and strip them off when matching.

.. This also pulls in a partial update to doc.go.

### DIFF
--- a/internal/services/frontend/doc.go
+++ b/internal/services/frontend/doc.go
@@ -117,7 +117,7 @@
 // This includes the blade ID number and associated URI.
 //
 // GET /api/racks/{rackid}/blades/{bladeid}
-// Returns the detail information for the blade numbed by bladeid in the rack
+// Returns the detail information for the blade number by bladeid in the rack
 // identified by rackid.
 //
 //
@@ -162,6 +162,9 @@
 // supplied document is invalid in some way, or if the supplied username is not
 // known.
 //
+// This operation is normally gated by an If-Match on a previously returned
+// revision ETag.
+//
 // POST /api/users/{username}
 // Creates a record for the user matching username, or an error if the supplied
 // document is invalid in some way. If the user record is successfully created,
@@ -174,9 +177,6 @@
 // If the username already exists and the supplied document differs from the
 // existing user record in any significant way, the response will be an HTTP
 // 409 (Conflict) status.
-//
-// This operation is normally gated by an If-Match on a previously returned
-// revision ETag.
 //
 // DELETE /api/users/{username}
 // Deletes the record for the user matching username, or an error is the
--- a/internal/services/frontend/doc.go
+++ b/internal/services/frontend/doc.go
@@ -10,6 +10,10 @@
 // specific source file associated with it.  These, and their associated
 // URL subtrees, are:
 //
+// frontend.go : all URIs
+// This is the starting point for the http processing.  It initializes the http
+// service attributes, including the initialization of the modules above.
+//
 // faults.go : /api/faults subtree
 // Implements the service handler to retrieve information on faults, to inject
 // faults into the simulated structure, to create and view devops actions.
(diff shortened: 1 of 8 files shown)


internal/services/frontend/doc.go
```diff
@@ -1,145 +1,170 @@
...
+// There are also a number of supporting files:
+//
+// frontend.go
+// Base file containing the global definition for the package and the main() entry point
+//
+// Service Syntax
+// ==============
+//
+// In the below definitions, the "{" and "}" characters surround an object name/key and are not included
+// in the URI. For example, to read the record of the user with the name "Bob", the URI would be
```
MikeCJohnsonUk, line 50: The URI would be ...?

I know it wasn't there in the original version, but still...

internal/services/frontend/doc.go
```diff
@@ -1,145 +1,170 @@
...
+//
+// If the supplied username is already known, and the supplied document exactly
+// matches the existing record, this will be interpreted as a duplicate request
+// and will result in an HTTP 200 (OK) status code. If the username already exists
+// and the supplied document differs from the existing user record in any significant
+// way, the response will be an HTTP 409 (Conflict) status.
+//
+// DELETE - /api/users/{username}
+// Deletes the record for the user matching username, or an error is the supplied
+// username is not known.
```
MikeCJohnsonUk, line 84: Presumably, an error would also be returned if the account was protected?

internal/services/frontend/doc.go
```diff
@@ -1,145 +1,170 @@
...
+//
+// PUT /api/users/{username}?op=[login|logout]
+// Updates the user record the for user matching username according to the supplied
+// operation, or returns and error if the supplied username is not known, or the
+// operation code is invalid in some way.
```
MikeCJohnsonUk, line 89: I suspect this is an old comment related to back when we had operations other than login/logout - enable/disable perhaps? Neither of the current ops should cause an update to the user record. Unless we add a last logged-in/out timestamp field that is.

internal/services/frontend/doc.go
```diff
@@ -1,145 +1,170 @@
...
+//
+// POST - /api/workloads/{workloadname}
+// Creates a record for the workload matching workloadname, or an error if the supplied
+// document is invalid in some way, or if the supplied workloadname is already known.
+//
+// DELETE - /api/workloads/{workloadname}
+// Deletes the record for the workload matching workloadname, or an error is the supplied
+// workloadname is not known.
+//
+// PUT /api/workloads/{workloadname}?op=[login|logout|enable|disable]
```
MikeCJohnsonUk, line 114: Does a workload need login / logout? I see the use for enable, disable and perhaps other operations e.g. start, stop, suspend etc, but it looks like the cut-n-paste fairy was a little too basic here ;-)

internal/services/frontend/doc.go
```diff
@@ -1,145 +1,170 @@
...
+//
+// GET - /api/stepper
+// Returns the current simulated time, mode and advance rate
+//
+// GET - /api/stepper/now
+// Returns the current simulated time
+//
+// PUT - /api/stepper?advance
+// Moves simulated time forward, so long as the mode is manual
+// This relies on an ETag returned by either of the two preceding GET operations
```
MikeCJohnsonUk, line 150: Since you bring it up, should there be mentions of the need, or not, for an ETag in the preceding PUT calls for other parts of the API namespace?

internal/services/frontend/doc.go
```diff
@@ -1,145 +1,170 @@
...
+// GET - /api/faults
+//
+// TODO
+//
+// /api/logs
+//
+// /api/faults
+// /api/inventory
+// will be a configuration file. you will be
+// /point to cmd. Few of them per server.
```
MikeCJohnsonUk, line 168: Not sure what the second sentence is trying to say.

internal/services/frontend/etag.go
```diff
@@ -0,0 +1,20 @@
...
+
+// formatAsEtag converts a number into a string formatted to meet the ETag
+// specification.
+func formatAsEtag(rev int64) string {
+	return fmt.Sprintf("\"%d\"", rev)
+}
+
+// parseAsMatchTag converts a string into the revision number that would have
+// been provided as a previous ETag, or an error if it is unable to parse it.
+func parseAsMatchTag(match string) (int64, error) {
```
MikeCJohnsonUk, line 17: Ok, I'm probably being over-picky here, but why "format as ETag" vs "parse as MatchTag". Shouldn't they both be either ETag or MatchTag?

MikeCJohnsonUk (commented):
I'll bet you wished you'd never touched doc.go ;-)

internal/services/frontend/doc.go
Jim3Things: Ok, I'll help with this and the other doc.go issues below - this was a partial update of doc.go inasmuch as I was in the middle of updating it when I it the ETag issue.  I got as far as 'There are also a number of supporting files:'

I should have said that everything after that is still on its way.  At least to avoid you reviewing them.  This is a good list, at least.

internal/services/frontend/etag.go
Jim3Things: I thought about that when writing it and could go either way.  I was triggering off the header field name it was associated with "ETag" goes out and "If-Match" comes back.  But since we know that "If-Match" uses ETag values, it probably makes more sense just to say parseAsETag.

internal/services/frontend/doc.go
```diff
@@ -40,131 +44,211 @@
...
+// supplied username is not known or is protected against deletion.
 //
+// PUT /api/users/{username}?op=[login|logout]
+// Updates the user record the for user matching username according to the
+// supplied operation, or returns and error if the supplied username is not
+// known, or the operation code is invalid in some way.
 //
-// /api/stepper
-// ------------
+// PUT /api/users/{username}?password
```
MikeCJohnsonUk, line 190: There are other fields that can be updated right? Such as canManageAccounts, enabled etc.

internal/services/frontend/doc.go
```diff
@@ -40,131 +44,211 @@
...
+//
+// GET /api/racks/{rackid}/blades
+// Returns the summary list of the blades in the rack identified by rackid.
+// This includes the blade ID number and associated URI.
+//
+// GET /api/racks/{rackid}/blades/{bladeid}
+// Returns the detail information for the blade numbed by bladeid in the rack
+// identified by rackid.
+//
+//
```
MikeCJohnsonUk, line 123: No mention of tors and/or pdus?

Note also, the update inventory store allows for multiple tors and pdus but at the moment the api to this defaults to a single pdu/tor with id=0

internal/services/frontend/doc.go
```diff
@@ -40,131 +44,211 @@
...
+// Returns the detail information about the rack identified by rackid.  This
+// includes information about the blades and rack-level support components,
+// such as the TOR and PDU.
+//
+// GET /api/racks/{rackid}/blades
+// Returns the summary list of the blades in the rack identified by rackid.
+// This includes the blade ID number and associated URI.
+//
+// GET /api/racks/{rackid}/blades/{bladeid}
+// Returns the detail information for the blade numbed by bladeid in the rack
```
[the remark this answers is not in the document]
Jim3Things: fixed.

internal/services/frontend/doc.go
Jim3Things: Yes, covered in the text starting at line 165.  But the change for those attributes does not include password, and changing the password does not include attributes other than the password.

internal/services/frontend/doc.go
Jim3Things: yes, to all.  Inventory.go just now does not support operations against pdu or tor specifically.  We should add them, but they're not there just now.

internal/services/frontend/doc.go
```diff
@@ -40,131 +44,211 @@
...
+// and will result in an HTTP 200 (OK) status code.
 //
-// GET - /api/racks/{rack-id}/blades/{blade-id} //Returns a record details of a single blade.
+// If the username already exists and the supplied document differs from the
+// existing user record in any significant way, the response will be an HTTP
+// 409 (Conflict) status.
 //
-// GET - /api/racks/{rack-id}/TOR //To get details about a specific blade in a specific rack
+// This operation is normally gated by an If-Match on a previously returned
+// revision ETag.
```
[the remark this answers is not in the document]
Jim3Things: good catch - this was meant to be on the PUT, not on the POST.  ETag check makes sense on an update, but not on a create.  I've moved it.

MikeCJohnsonUk (approved):

PR #204, opened 2021-01-25
Moving decode to internal/common

manpreet85:
26jan21

Created new branch.

Cut the two files and pasted into internal/common.


Jim3Things (changes requested):
You need to change the package in the files.

Did you build the system after making the change?

manpreet85:
Please help, i got 157 Problems after i changed the package to common.

If i build, would the problems disappear? Please advise

Jim3Things:
I'd suggest a make clean followed by a make build -j (after starting etcd), and see where that gets you.  Changing the package to match the package of common in these two files should not cause a bunch of new problems.

Jim3Things (approved):

PR #205, opened 2021-01-29
Add simulation status reporting

Jim3Things:


internal/services/frontend/httpErrors.go
```diff
@@ -11,6 +11,7 @@ import (
 	"strconv"
 
 	"github.com/Jim3Things/CloudChamber/internal/tracing"
+	err2 "github.com/Jim3Things/CloudChamber/pkg/errors"
```
MikeCJohnsonUk, line 14: I found myself reacting to the "err2", the '2' part in particular.

Would perhaps just "e" suffice here?

internal/services/frontend/session_manager.go
```diff
@@ -85,6 +106,24 @@ func removeSession(session *sessions.Session) {
...
+// use by the caller.
+func removeSessionById(id int64) (sessionState, bool) {
+	mutex.Lock()
+	defer mutex.Unlock()
+
+	purgeStaleSessions()
+
+	if entry, ok2 := activeSessions[id]; ok2 {
+		delete(activeSessions, id)
+		delete(timeouts, entry.timeout)
```
MikeCJohnsonUk, line 120: Should these two lines be commoned up with those in purgeStaleSessions() into a deleteSession() func to ensure the same delete actions are always taken for a session removal?

internal/services/frontend/session_manager.go
```diff
@@ -116,6 +155,20 @@ func getSession(session *sessions.Session) (sessionState, bool) {
...
+
+	purgeStaleSessions()
+
+	entry, ok := activeSessions[id]
+	return entry, ok
+}
+
 // purgeStaleSessions removes stale sessions from the active session list.  It
 // assumes that the number of sessions is not so large that the active session
 // list cannot be scanned.
```
MikeCJohnsonUk, line 174: Do we have anything anywhere which limits the number of active sessions? Should we? Just to stop the runaways.

internal/services/frontend/simulation_test.go
```diff
@@ -0,0 +1,144 @@
...
+	request = httptest.NewRequest("GET", status.Sessions[0].Uri, nil)
+	request.Header.Set("Content-Type", "application/json")
+
+	response = ts.doHTTP(request, response.Cookies())
+	require.Equal(http.StatusForbidden, response.StatusCode)
+}
+
+func TestSimulationTestSuite(t *testing.T) {
+	suite.Run(t, new(SimulationTestSuite))
+}
```
MikeCJohnsonUk, line 144: Do we need anything to cover non-admin sessions. non-admin calls and multiple active sessions?

internal/services/frontend/httpErrors.go
Jim3Things: I'll try that.

internal/services/frontend/session_manager.go
Jim3Things: We don't.  I'll add an issue.

internal/services/frontend/simulation_test.go
Jim3Things: Yeah, should do.  Forgot that when doing the tests.

internal/services/frontend/session_manager.go
Jim3Things: Hmm.  Yeah, not much of a gain, but yeah, it does replace some repeat code.

I also realize that there's a bug here - we do the wrong thing on timeout collisions.  I'm not sure we'd see such a collision, but I should fix that anyway.

MikeCJohnsonUk: I wasn't thinking about the savings per-se, more that we have a consistent set of operations. I've seen other project run afoul of that.

But the savings, are nice too.

internal/services/frontend/session_manager.go
```diff
@@ -214,58 +199,229 @@ func dumpSessionState(session *sessions.Session) string {
...
 		}
 	}
 
-	user, _, err := dbUsers.Read(ctx, entry.name)
-	return user, err
+	st.lastID++
+	entry.timeout = time.Now().Add(st.maxInactive)
+	st.addToTables(st.lastID, entry)
+
+	return st.lastID, nil
```
MikeCJohnsonUk, line 273: Possibly over picky, but when exactly does the deferred unlock happen wrt the acquisition of the current value of st.lastID? 

I think this is ok, but just wanted to check.

internal/services/frontend/session_manager.go
```diff
@@ -214,58 +199,229 @@ func dumpSessionState(session *sessions.Session) string {
...
+// Note that expired entries are first removed from the sessionTable prior to
+// processing this request.
+func (st *sessionTable) touch(id int64) (sessionState, bool) {
+	st.m.Lock()
+	defer st.m.Unlock()
+
+	st.purge()
+
+	if entry, ok := st.known[id]; ok {
+		st.deleteFromTimeoutList(id, entry.timeout)
```
MikeCJohnsonUk, line 321: being picky again: removeFromTimeoutList() ?

internal/services/frontend/session_manager.go
Jim3Things: According to go, defer executes after the return, so we should be good.

internal/services/frontend/session_manager.go
Jim3Things: yeah, that's pretty picky :).  Done.

MikeCJohnsonUk (approved):

PR #208, opened 2021-02-03
Merge observer GUI into repo

Jim3Things:
This is the second checkin in the project restructuring.

This moves the typescript-based GUI into the base repo, and adjusts the install step of the core build to use the new location.  No other functional or documentation changes are made.


clients/observer/Makefile
```diff
@@ -0,0 +1,24 @@
+PROJECT = $(GOPATH)/src/github.com/Jim3Things/CloudChamber
+
+PROJECT_UI = ../cloud_chamber_react_ts
```
MikeCJohnsonUk, line 3: If we still need this file, then this variable is getting set to the wrong value.

Thinking further, I suspect what we'll need is a Makefile.inc which contains all the common variable and command definitions. No, I'm not suggesting you do that right now...

MikeCJohnsonUk (approved):
I'm basically assuming all the src files are just being moved without change. The only possible issue I could see was in the original Makefile which likely has a stale value for the PROJECT_UI definition, but I think we'll need to revamp all that stuff anyway so I wouldn't worry about it for now.

PR #209, opened 2021-02-05
Add simulation status and trace retrieval cmdlets

Jim3Things:
This adds cmdlets to retrieve the current simulation status, and to forcibly clear out logged in sessions.  It also adds cmdlets to retrieve the trace policy and extracts of the current traces.

Note that the protos definitions are uncommented as they are extracts from the protos and will soon be replaced by generated code.

### DIFF
--- a/clients/powershell/Cmdlets/script_tests/Simulation.Tests.ps1
+++ b/clients/powershell/Cmdlets/script_tests/Simulation.Tests.ps1
@@ -1,0 +1,36 @@
+﻿Describe "Status checks against the simulation" {
+    BeforeAll {
+        Import-Module ..\bin\Debug\netstandard2.0\CloudChamber.Cmdlets.dll
+    }
+
+    BeforeEach {
+        $sess = Connect-CcAccount -ClusterUri http://localhost:8080 -Name admin -Password adminPassword
+    }
+
+    AfterEach {
+        Disconnect-CcAccount -Session $sess
+    }
+
+    It "Gets the simulation status" {
+        $status = Get-CcCluster -Session $sess
+        $span = New-TimeSpan -Hours 1
+        $status.InactivityTimeout | Should Be $span
+
+        $now = Get-Date
+        ($status.FrontEndStartedAt -lt $now) | Should Be $true
+    }
+
+    It "Gets the summary list of current active sessions" {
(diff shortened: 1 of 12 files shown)


clients/powershell/Cmdlets/script_tests/Simulation.Tests.ps1
```diff
@@ -0,0 +1,36 @@
+﻿Describe "Status checks against the simulation" {
+    BeforeAll {
+        Import-Module ..\bin\Debug\netstandard2.0\CloudChamber.Cmdlets.dll
+    }
+
+    BeforeEach {
+        $sess = Connect-CcAccount -ClusterUri http://localhost:8080 -Name admin -Password adminPassword
```
MikeCJohnsonUk, line 7: It always makes me nervous seeing passwords in files. I know it's necessary sometimes, but old habits die hard!

clients/powershell/Cmdlets/script_tests/Traces.Tests.ps1
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    AfterEach {
+        Disconnect-CcAccount -Session $sess
+    }
+
+    It "Gets the current trace policy" {
+        $policy = Get-CcTracePolicy -Session $sess
+        $policy.MaxEntriesHeld | Should Be 1000
+        $policy.FirstId | Should BeGreaterThan -2
```
MikeCJohnsonUk, line 17: Are we using -1 as a signal somewhere?

clients/powershell/Cmdlets/script_tests/Simulation.Tests.ps1
Jim3Things: Indeed.  But this name and password is also in the config for the test clusters...

Hopefully we'll be able to move on from passwords before long.

clients/powershell/Cmdlets/script_tests/Traces.Tests.ps1
Jim3Things: Yes, -1 means 'the start of all history'.  Since we don't know what traces have been made for the test cluster the tests we can do are pretty minimal...

MikeCJohnsonUk (approved):

PR #210, opened 2021-02-06
Add primitives to store inventory (definitions only at present) in the persisted store

MikeCJohnsonUk:
This change adds the primitives required to save and retrieve
descriptions of various items of inventory in the persistent
store. The goal of the set of routines is to abstract the store record
formats and naming conventions from the clients of the persistent
inventory and to provide a simple interface to access and update those
records.

The update includes definitions and primitives to extend the inventory
items to include regions and zones along with racks, pdus, tors and
blades.

For examples on expected usage patterns, see

definition_test.go\createInventory()
definition_test.go\TestRackFetchChildren() 

Notes:

config\InventoryReader*.go was moved to clients\inventory\reader*.go
along with the associated test and default config yaml files.

The loading of the inventory from the configuration files is not
currently being persisted in the store. (Future PR)

Although there is a definition of an interface to the inventory items,
it is by no means clear that it will actually be required and may
well be removed. At present, if nothing else, it serves as a
description of the routines in the inventory package.

The <class>.Delete() functions work as if the force option was always
set. This means that at present any attempt to delete a non-existent
record will report success.

The <class>.Delete() functions need to be updated to use the currently
missing DeleteMultiple() call to delete the object and any associated
index record within the same transaction. (Future PR)


internal/clients/inventory/inventory.go
```diff
@@ -0,0 +1,881 @@
...
+// are not set.
+//
+type inventoryItem interface {
+	// Use to set the attribues of an object within the inventory
+	//
+	SetDetails(ctx context.Context, details *interface{})
+	GetDetails(ctx context.Context) *interface{}
+	GetRevision(ctx context.Context) int64
+	GetRevisionRecord(ctx context.Context) int64
+	GetRevisionStore(ctx context.Context) int64
```
Jim3Things, line 650: I'm not sure, but if you go with splitting the revision fields out into their own struct in the earlier file, you may need to have this split out into its own interface.  I _think_ you would then include that interface as an unnamed entry here - but that is a guess.

internal/clients/inventory/inventory.go
```diff
@@ -0,0 +1,881 @@
...
+	NewTor(ctx context.Context,   name string) (*interface{}, error)
+	NewBlade(ctx context.Context, name string) (*interface{}, error)
+
+	ListPdus(ctx context.Context)   (int64, *map[int64]*interface{}, error)
+	ListTors(ctx context.Context)   (int64, *map[int64]*interface{}, error)
+	ListBlades(ctx context.Context) (int64, *map[int64]*interface{}, error)
+
+	FetchPdus(ctx context.Context)   (int64, *map[int64]*interface{}, error)
+	FetchTors(ctx context.Context)   (int64, *map[int64]*interface{}, error)
+	FetchBlades(ctx context.Context) (int64, *map[int64]*interface{}, error)
```
Jim3Things, line 738: It's been feeling like there's a collection type in here somewhere, but I've not had the a-ha moment on that yet....

internal/clients/inventory/reader.go
```diff
@@ -190,25 +190,27 @@ func ReadInventoryDefinitionFromFile(ctx context.Context, path string) (*pb.Regi
 }
 
 // toDefinitionRegionInternal converts intermediate values to the final format
-// One important difference is that the intermediate is array based.
+// One important differnce is that the intermediate is array based.
```
Jim3Things, line 193: Uhm, I'd undo this change... :)

pkg/errors/errors.go
```diff
@@ -376,143 +394,190 @@ func (eutvsa ErrUnableToVerifySystemAccount) Error() string {
...
 //
 type ErrRackAlreadyExists struct {
+	Region string
 	Zone string
 	Rack string
 }
 
 func (erae ErrRackAlreadyExists) Error() string {
-	return fmt.Sprintf("CloudChamber: rack %q in zone %q already exists", erae.Rack, erae.Zone)
+	return fmt.Sprintf("CloudChamber: rack %q in zone %q already exists in region %q", erae.Rack, erae.Zone, erae.Region)
```
Jim3Things, line 441: Update the xxxAddress functions in the errors package and use them?

pkg/errors/errors.go
```diff
@@ -376,143 +394,190 @@ func (eutvsa ErrUnableToVerifySystemAccount) Error() string {
...
+	Rack string
+	Tor int64
 }
 
-func (ebae ErrBladeAlreadyExists) Error() string {
-	return fmt.Sprintf(
-		"CloudChamber: %s already exists",
-		bladeAddress(ebae.Zone, ebae.Rack, ebae.Blade))
+func (etae ErrTorAlreadyExists) Error() string {
+	return fmt.Sprintf("CloudChamber: tor %v in region %q, zone %q, rack %q already exists", etae.Tor, etae.Region, etae.Zone, etae.Rack)
```
Jim3Things, line 538: I'm not finding all the cases, but shouldn't this use %d rather than %v?

pkg/protos/inventory/internal.proto
```diff
@@ -1,12 +1,98 @@
...
+// internal, universal representation of the various items in the inventory. All
+// other forms can be converted into the internal format and likewise, all the
+// other forms can be generated from the internal format.
 
 syntax = "proto3";
 
 package inventory;
 
+import "github.com/Jim3Things/CloudChamber/pkg/protos/inventory/common.proto";
+import "github.com/Jim3Things/CloudChamber/pkg/protos/inventory/actual.proto";
```
Jim3Things, line 11: It's too early to matter, but I'm pretty sure that observed will have some states that differ from actual - e.g. observed will have a 'in suspicion' state, where it has been too long since the item was seen, while actual will not.  Not sure treating them as the same type will work, but it is probably too early to tell.

Jim3Things (commented):
I mostly found some simplification points.

internal/clients/inventory/definition.go
```diff
@@ -0,0 +1,2574 @@
...
+
+// SetDetails is used to attach some attribute information to the object.
+//
+func (r *Root) SetDetails(ctx context.Context, details *pb.RootDetails) {
+
+	r.details  = details
+	r.revision = store.RevisionInvalid
+}
+
+// GetDetails is a
```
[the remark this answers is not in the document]
MikeCJohnsonUk: Fixed

internal/clients/inventory/definition.go
```diff
@@ -0,0 +1,2574 @@
...
+	return region, nil
+}
+
+// ListChildren is a
+//
+func (r *Root) ListChildren(ctx context.Context) (int64, []string, error) {
+	
+	records, rev, err := r.Store.List(ctx, store.KeyRootInventory, r.KeyChildIndex)
+
+	if err == errors.ErrStoreIndexNotFound(r.KeyChildIndex) {
```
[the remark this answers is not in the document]
MikeCJohnsonUk: I'll leave this for a future decision/change.

internal/clients/inventory/definition.go
```diff
@@ -0,0 +1,2574 @@
...
+//
+func (r *Region) NewChild(ctx context.Context, name string) (*Zone, error) {
+
+	z, err := NewZone(ctx, r.Store, r.Table, r.Region, name)
+
+	if err != nil {
+		return nil, err
+	}
+
+	return z, err
```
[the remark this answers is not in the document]
MikeCJohnsonUk: Fixed to just return NewZone() and so on for the other types.

internal/clients/inventory/definition.go
```diff
@@ -0,0 +1,2574 @@
...
+//
+type Root struct {
+
+	Store         *store.Store
+	KeyChildIndex  string
+	Table          string
+
+	revision       int64
+	revisionRecord int64
+	revisionStore  int64
```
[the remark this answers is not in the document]
MikeCJohnsonUk: Done.

internal/clients/inventory/definition.go
```diff
@@ -0,0 +1,2574 @@
...
+// from the underlying store, or as a staging area in preparation for updates
+// to the store.
+//
+type Region struct {
+
+	Store          *store.Store
+	KeyChildIndex  string
+	KeyIndexEntry  string
+	Key            string
+	Table          string
```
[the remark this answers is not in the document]
MikeCJohnsonUk: Sigh. It turns out, there's a lot of variation one a similar theme here. There are only two places where you get all four. A more common grouping is KeyIndexEntry, Key, Table so something should be possible here. I think I want to think on this a little more before making any changes.

internal/clients/inventory/inventory.go
MikeCJohnsonUk: Yeah, me too.

internal/clients/inventory/reader.go
MikeCJohnsonUk: Sigh. Fixed.

pkg/errors/errors.go
MikeCJohnsonUk: Done

pkg/errors/errors.go
MikeCJohnsonUk: Fixed a bunch, can't find anymore, at least not in this module.

pkg/protos/inventory/internal.proto
MikeCJohnsonUk: Probably. We already have two "state" type enums, and I'd kinda prefer to not add another unless it's necessary for these type of dynamic, non-persisted states/conditions.

internal/clients/inventory/definition.go
```diff
@@ -0,0 +1,2574 @@
...
+		r.KeyIndexEntry : r.Region,
+		r.Key           : v,
+	}
+
+	rev, err := r.Store.CreateMultiple(ctx, store.KeyRootInventory, keySet)
+
+	switch (err) {
+	case nil:
+	case errors.ErrStoreAlreadyExists(r.KeyIndexEntry): return store.RevisionInvalid, errors.ErrRegionAlreadyExists(r.Region)
+	case errors.ErrStoreAlreadyExists(r.Key):           return store.RevisionInvalid, errors.ErrRegionAlreadyExists(r.Region)
```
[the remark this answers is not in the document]
MikeCJohnsonUk: This got "fixed" with a mapErrorXxxx() function.so no longer a problem.

internal/clients/inventory/definition.go
```diff
@@ -0,0 +1,2574 @@
...
+
+	rev, err := r.Store.Update(ctx, store.KeyRootInventory, r.Key, revUpdate, v)
+
+	if err == errors.ErrStoreKeyNotFound(r.Key) {
+		return store.RevisionInvalid, errors.ErrRegionNotFound(r.Region)
+	}
+
+	if err != nil {
+		return store.RevisionInvalid, err
+	}
```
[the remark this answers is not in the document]
MikeCJohnsonUk: Yup, I did something like this and it works out fairly cleanly. Having a struct specific function works well as then all the field names for the new error are readily available.

Take a look at the new variants and see if that works for you too.

internal/clients/inventory/definition.go
```diff
@@ -0,0 +1,2574 @@
...
+
+	rev, err := z.Store.CreateMultiple(ctx, store.KeyRootInventory, keySet)
+
+	switch (err) {
+	case nil:
+	case errors.ErrStoreAlreadyExists(z.KeyIndexEntry): return store.RevisionInvalid, errors.ErrZoneAlreadyExists{Region: z.Region, Zone: z.Zone}
+	case errors.ErrStoreAlreadyExists(z.Key):           return store.RevisionInvalid, errors.ErrZoneAlreadyExists{Region: z.Region, Zone: z.Zone}
+	default:                                            return store.RevisionInvalid, err
+	}
```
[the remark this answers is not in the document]
MikeCJohnsonUk: Actually, I have something which I think is much simpler, since we observe that for the nil and default cases, the mapXxx() just returns the existing err for nil and default, and then a simple if A || B return mappedError for the mapped cases.

Take a look at the new variants and see if that works for you too.

internal/clients/inventory/definition.go
```diff
@@ -0,0 +1,2574 @@
...
+//
+func (r *Rack) NewPdu(ctx context.Context, ID int64) (*Pdu, error) {
+
+	p, err := NewPdu(ctx, r.Store, r.Table, r.Region, r.Zone, r.Rack, ID)
+
+	if err != nil {
+		return nil, err
+	}
+
+	return p, err
```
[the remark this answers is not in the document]
MikeCJohnsonUk: Agreed. Done.

internal/clients/inventory/inventory.go
```diff
@@ -0,0 +1,881 @@
...
+	keyFormatIndexEntryPdu     = keyFormatIndexPdus    + "%d"
+	keyFormatIndexEntryTor     = keyFormatIndexTors    + "%d"
+	keyFormatIndexEntryBlade   = keyFormatIndexBlades  + "%d"
+
+	keyFormatRegion = "%s/data/" + prefixRegion + "/%s"
+	keyFormatZone   = "%s/data/" + prefixRegion + "/%s/" + prefixZone + "/%s"
+	keyFormatRack   = "%s/data/" + prefixRegion + "/%s/" + prefixZone + "/%s/" + prefixRack + "/%s"
+	keyFormatPdu    = "%s/data/" + prefixRegion + "/%s/" + prefixZone + "/%s/" + prefixRack + "/%s/" + prefixPdu   + "/%v"
+	keyFormatTor    = "%s/data/" + prefixRegion + "/%s/" + prefixZone + "/%s/" + prefixRack + "/%s/" + prefixTor   + "/%v"
+	keyFormatBlade  = "%s/data/" + prefixRegion + "/%s/" + prefixZone + "/%s/" + prefixRack + "/%s/" + prefixBlade + "/%v"
```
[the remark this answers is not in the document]
MikeCJohnsonUk: Done

internal/clients/inventory/inventory.go
MikeCJohnsonUk: I believe so. At least it's what I've done and it seems to be working.

internal/clients/inventory/definition.go
```diff
@@ -0,0 +1,2574 @@
...
+
+	names := make([]int64, 0, len(*records))
+
+	for k, v := range *records {
+	
+		name := strings.TrimPrefix(k, r.KeyIndexPdu)
+
+		// Verify that the "index" part of the name is numeric
+		//
+		intName, err := strconv.ParseInt(name, 10, 0)
```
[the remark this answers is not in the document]
MikeCJohnsonUk: Done

internal/clients/inventory/definition.go
```diff
@@ -0,0 +1,2574 @@
...
+	case nil:
+	case errors.ErrStoreAlreadyExists(r.KeyIndexEntry): return store.RevisionInvalid, errors.ErrRegionAlreadyExists(r.Region)
+	case errors.ErrStoreAlreadyExists(r.Key):           return store.RevisionInvalid, errors.ErrRegionAlreadyExists(r.Region)
+	default:                                            return store.RevisionInvalid, err
+	}
+
+	r.record         = record
+	r.revision       = rev
+	r.revisionRecord = rev
+	r.revisionStore  = rev
```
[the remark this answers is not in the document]
MikeCJohnsonUk: Agreed. Done.

internal/clients/inventory/definition.go
```diff
@@ -0,0 +1,2574 @@
...
+// revision information using the information returned by the store
+// operation.
+//
+func (r *Rack) Delete(ctx context.Context, unconditional bool) (int64, error) {
+
+	var revDelete = r.revisionRecord
+
+	if unconditional == true {
+		revDelete = store.RevisionInvalid
+	}
```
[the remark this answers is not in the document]
MikeCJohnsonUk: Turns out there is something, so done.

internal/clients/inventory/definition.go
```diff
@@ -0,0 +1,2376 @@
...
+func (r *revisionInfo) GetRevisionRecord(ctx context.Context) int64 {
+	return r.revisionRecord
+}
+
+// GetRevisionStore returns the revison of the underlying store ifself as 
+// determined at the time of the last Create() Read() for the object. The
+// store revision is not reset by a SetDetails() call and is provided 
+// for information only.
+//
+func (r *revisionInfo) GetRevisionStore(ctx context.Context) int64 {
```
Jim3Things, line 290: For these new functions, where you're passing context.Context but not using it, it would be best if the parameter was named "_"...

internal/clients/inventory/definition.go
```diff
@@ -0,0 +1,2376 @@
...
+	return err
+}
+
+func (r *Region) mapErrStoreAlreadyExists(err error) error {
+	if err == errors.ErrStoreAlreadyExists(r.KeyIndexEntry) || err == errors.ErrStoreAlreadyExists(r.Key) {
+		return errors.ErrRegionAlreadyExists{Region: r.Region}
+	}
+	
+	return err
+}
```
Jim3Things, line 770: Out curiousity, why not have a single map function that handles all the transformations - that way you can add new transforms as needed without changing the caller...

pkg/errors/errors.go
```diff
@@ -815,3 +914,169 @@ type ErrSessionNotFound int64
...
+	Name            string
+	ActualTable     string
+	DefinitionTable string
+	ObservedTable   string
+	TargetTable     string
+}
+
+func (e ErrTableNameInvalid) Error() string {
+	return fmt.Sprintf(
+		"CloudChamber: table name %q is not one of the valid options [%q, %q, %q, %q",
```
Jim3Things, line 930: Missing closing ']'

Jim3Things (commented):
a couple of minor comments

internal/clients/inventory/definition.go
MikeCJohnsonUk: I debated whether or not to be passing ctx at all. They're really just accessor funcs and it was only a reflex to add the ctx "just-in-case".

I think I'd be completely ok with removing ctx for these simple routines. It's hard to see why it would ever be useful. Thoughts?

Jim3Things: I'd remove the parameter then.

MikeCJohnsonUk: Removed.

pkg/errors/errors.go
MikeCJohnsonUk: Done

internal/clients/inventory/definition.go
MikeCJohnsonUk: Good question.

I've now done this and it even simplifies some of the rack error mappings.


However, while fixing this, I spotted a few problems associated with the Delete() methods for all six cases.

1. Most of the methods fail to delete the index key for the record
2. All of the methods assume the store returning them an error indicating a problem with the "index". The store generates no such error. What it will do is report a no suck key error which the inventory code should recognize as applying to the index value and as such generate the index error for the object.
3. The index error can/should be enhanced to include the full object address as appropriate, just like the not found error.

None of these problems are critical and won't case problems other that possibly leading to the wrong error being reported, or leaving around some orphan index keys while a test runs. The overall test cleanup will remove these orphans.

Given that my next hotly approaching change is to add the proper atomic index delete code currently absent due to the missing store.DeleteMultiple() function, I'm proposing that I flush out these three issues and fix them in this subsequent change.

Jim3Things (approved):

PR #211, opened 2021-02-06
Add start of C# support in protos

Jim3Things:
Also fixed two failures: one in the powershell unit tests, and one in ETag handling in the web UI.  Both found while validating that the new protos were json consistent with the old ones.

### DIFF
--- a/clients/observer/src/proxies/Session.tsx
+++ b/clients/observer/src/proxies/Session.tsx
@@ -109,7 +109,8 @@ export function getETag(resp: Response): number {
         return -1
     }
 
-    return parseInt(tag, 10)
+    const value = tag.replace("\"", "")
+    return parseInt(value, 10)
 }
 
 // Set the ETag into a header as a match condition
--- a/clients/powershell/Cmdlets/script_tests/Simulation.Tests.ps1
+++ b/clients/powershell/Cmdlets/script_tests/Simulation.Tests.ps1
@@ -16,8 +16,8 @@
         $span = New-TimeSpan -Hours 1
         $status.InactivityTimeout | Should Be $span
 
-        $now = Get-Date
-        ($status.FrontEndStartedAt -lt $now) | Should Be $true
+        $now = (Get-Date).ToUniversalTime()
+        $status.FrontEndStartedAt | Should BeLessThan $now
     }
 
     It "Gets the summary list of current active sessions" {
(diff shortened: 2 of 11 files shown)


clients/observer/src/proxies/Session.tsx
```diff
@@ -109,7 +109,8 @@ export function getETag(resp: Response): number {
         return -1
     }
 
-    return parseInt(tag, 10)
+    const value = tag.replace("\"", "")
```
MikeCJohnsonUk, line 112: These are our own tag values right? Any issues with unexpected placements of embedded " characters? Is removal always the right thing to do? I mean, it probably is, but...

clients/powershell/Cmdlets/script_tests/Simulation.Tests.ps1
```diff
@@ -16,8 +16,8 @@
         $span = New-TimeSpan -Hours 1
         $status.InactivityTimeout | Should Be $span
 
-        $now = Get-Date
-        ($status.FrontEndStartedAt -lt $now) | Should Be $true
+        $now = (Get-Date).ToUniversalTime()
+        $status.FrontEndStartedAt | Should BeLessThan $now
```
MikeCJohnsonUk, line 20: I believe it, but could you indicate why the original didn't work? Or is this just stylistic?

clients/observer/src/proxies/Session.tsx
Jim3Things: They are.  This is the fallout of needing to wrap them in " characters.  Since they should be a number, this should work.  I can do a more thorough parse of them, but I'd like to hold off, as I think I'll want to just change the ETag handling all up to be an uninterpreted moniker, which is a bigger change and would avoid the parsing and formatting.

clients/powershell/Cmdlets/script_tests/Simulation.Tests.ps1
Jim3Things: It's both.  The bug was that Get-Date returns local time, the FrontEndStartAt is UTC, so the initial comparison was missing the timezone offset (and new clusters were failing because they hadn't been up for 8 hours).  Then, I didn't like the -lt check that was there, as it kept the error case from showing what the actual data was.  That was fixed by changing to this format.

MikeCJohnsonUk (approved):

PR #212, opened 2021-02-08
Use generated .cs protobuf files

Jim3Things:
Modifies the makefile to generate .cs files from a subset of the protobufs.

Modifies the cmdlets csproj to link to them, and to require the protobuf runtime.

The cmdlets are modified to use these new definitions, tests are modified when required.

The temporary data structure definitions, and the Newtonsoft json support are removed.


Jim3Things:
Two notes on this change: First, I expect to hold this until after Mike's change is through, just to limit conflict resolution.  Second, the cmdlets currently emit some XxxResponse structures that arrive over the wire.  Not sure if we should convert to a local type and return that.  Open to suggestions.

MikeCJohnsonUk (approved):

PR #216, opened 2021-02-10
Move cloudchamber simulation service down one directory level

Jim3Things:
+ gofmt project
+ remove some duplicate/stale vendor projects


MikeCJohnsonUk (approved):
Given this affects 11323 files, of which the PR page will only list the first 3000, I don't think there's anything meaning file I can do with this other than approve the change.

Fingers crossed!

PR #217, opened 2021-02-12
Add inventory cmdlets

Jim3Things:


clients/powershell/Cmdlets/script_tests/Inventory.Tests.ps1
```diff
@@ -0,0 +1,75 @@
...
+        }
+        $found | Should Be $true
+
+        $list.MaxBladeCount | Should Be 8
+        $capacity = $list.MaxCapacity
+        $capacity.Cores | Should Be 32
+        $capacity.MemoryInMb | Should Be 16834
+        $capacity.DiskInGb | Should Be 240
+        $capacity.NetworkBandwidthInMbps | Should Be 2048
+        $capacity.Accelerators.Count | Should Be 0
```
MikeCJohnsonUk, line 35: Perhaps I missed this, but where do these number come from?

Do we have a standard test inventory?

clients/powershell/Cmdlets/script_tests/Inventory.Tests.ps1
```diff
@@ -0,0 +1,75 @@
+﻿Describe "Operations against inventory" {
+    BeforeAll {
+        Import-Module ..\bin\Debug\netstandard2.0\CloudChamber.Cmdlets.dll
+    }
+
+    BeforeEach {
+        $sess = Connect-CcAccount -ClusterUri http://localhost:8080 -Name admin -Password adminPassword
```
MikeCJohnsonUk, line 7: Does this need to use the admin account?

clients/powershell/Cmdlets/script_tests/Inventory.Tests.ps1
Jim3Things: I've been using the standard inventory.yaml file that we use in our build.

clients/powershell/Cmdlets/script_tests/Inventory.Tests.ps1
Jim3Things: I used it because it is the only one we're sure to have.  I can create a temporary non-admin account in a BeforeAll action and delete it in an AfterAll action.  That may be better.

clients/powershell/Cmdlets/script_tests/Inventory.Tests.ps1
MikeCJohnsonUk: Fair enough. Probably don't have all the bits yet for anything better.

clients/powershell/Cmdlets/script_tests/Inventory.Tests.ps1
MikeCJohnsonUk: Your call. We probably should have a set of standard inventory / users/ config, but since we don't currently have all the necessary bit to consume those just yet...

MikeCJohnsonUk (approved):

PR #219, opened 2021-02-15
Convert  inventory definition tests to be based on the testify test suite package

MikeCJohnsonUk:
In preparation for additional tests, convert inventory definitions tests to be based on the testify package, as are the existing reader tests.

Requires moving to a common core test suite to allow each sub-package (reader, definition) to make use of the same common test setup core. 

Since the flag package must only parse flags exactly once to avoid issues, need to separate the flag parsing into a guaranteed run at most once section.

### DIFF
--- a/simulation/internal/clients/inventory/TestSuiteCore.go
+++ b/simulation/internal/clients/inventory/TestSuiteCore.go
@@ -29,19 +29,6 @@ type testSuiteCore struct {
 	store *store.Store
 }
 
-func runOnceRequired() bool{
-	lock.Lock()
-	defer lock.Unlock()
-
-	if initialized {
-		return false
-	}
-
-	initialized = true
-
-	return true
-}
-
 func runOnce() (*config.GlobalConfig, error) {
 	lock.Lock()
 	defer lock.Unlock()
--- a/simulation/internal/clients/inventory/TestSuiteCore.go
+++ b/simulation/internal/clients/inventory/TestSuiteCore.go
@@ -1,0 +1,93 @@
+package inventory
+
+import (
+	"flag"
+	"sync"
+
+	"github.com/stretchr/testify/suite"
+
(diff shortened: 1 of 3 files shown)


simulation/internal/clients/inventory/TestSuiteCore.go
```diff
@@ -0,0 +1,93 @@
...
+
+	if initialized {
+		return false
+	}
+
+	initialized = true
+
+	return true
+}
+
```
Jim3Things, line 44: What is this used for?  I didn't see a reference, but could have missed it.

simulation/internal/clients/inventory/reader_test.go
```diff
@@ -4,279 +4,264 @@ package inventory
...
-
-	_ = utf.Open(t)
-	defer utf.Close()
+func (ts *readerTestSuite) TestReadInventoryDefinition() {
+	assert := ts.Assert()
+	require := ts.Require()
 
 	response, err := ReadInventoryDefinition(context.Background(), "./testdata/Basic")
-	require.Nil(t, err)
+	require.Nil(err)
```
Jim3Things, line 45: If you're modifying these anyway, move to require.NoError?

simulation/internal/clients/inventory/TestSuiteCore.go
MikeCJohnsonUk: Well huh. I *was* using it, but no longer it seems.

Removed.

simulation/internal/clients/inventory/reader_test.go
MikeCJohnsonUk: You know that time when you pull on that little dangling thread...

Fixed. And a bunch of others.

Jim3Things (approved):

PR #220, opened 2021-02-16
Add ability to read inventory definition using an extended syntax

MikeCJohnsonUk:
Add the ability to load the inventory from a file using an extended syntax to cater for regions, zones and ports with an explicit wiring designation.

This syntax supersedes the current format and will shortly replace it in a forthcoming PR.

Remove superfluous x.Verify methods and replace calls with updated x.Validate() methods

### DIFF
--- a/simulation/internal/clients/inventory/reader.go
+++ b/simulation/internal/clients/inventory/reader.go
@@ -559,7 +559,7 @@ func toDefinitionRoot(xfrRoot *rootEx) (*pb.Definition_Root, error) {
 			Name: xfrRoot.Details.Name,
 			Notes: xfrRoot.Details.Notes,
 		},
-		Regions: make(map[string]*pb.Definition_Region),
+		Regions: make(map[string]*pb.Definition_Region, len(xfrRoot.Regions)),
 	}
 
 	// For each rack in the supplied configuration, create rack in the
@@ -573,10 +573,10 @@ func toDefinitionRoot(xfrRoot *rootEx) (*pb.Definition_Root, error) {
 			}
 		}
 		var (
-			ok		   bool
+			ok         bool
 			state      pb.State
 			condition  pb.Condition
-			hwType	   pb.Hardware_HwType
+			hwType     pb.Hardware_HwType
 			bootSource pb.BladeBootInfo_Method
 		)
 
@@ -595,7 +595,7 @@ func toDefinitionRoot(xfrRoot *rootEx) (*pb.Definition_Root, error) {
 				Location: xfrRegion.Details.Location,
 				Notes:    xfrRegion.Details.Notes,
 			},
-			Zones:   make(map[string]*pb.Definition_Zone),
+			Zones:   make(map[string]*pb.Definition_Zone, len(xfrRegion.Zones)),
 		}
 	
 		// Iterate over the set of zones in the supplied configuration
@@ -626,7 +626,7 @@ func toDefinitionRoot(xfrRoot *rootEx) (*pb.Definition_Root, error) {
 					Location: xfrZone.Details.Location,
 					Notes:    xfrZone.Details.Notes,
 				},
-				Racks: make(map[string]*pb.Definition_Rack),
+				Racks: make(map[string]*pb.Definition_Rack, len(xfrZone.Racks)),
 			}
 
 			// Iterate over the set of racks in the supplied configuration
@@ -659,9 +659,9 @@ func toDefinitionRoot(xfrRoot *rootEx) (*pb.Definition_Root, error) {
 						Location:  xfrRack.Details.Location,
 						Notes:     xfrRack.Details.Notes,
 					},
-					Pdus:   make(map[int64]*pb.Definition_Pdu),
-					Tors:   make(map[int64]*pb.Definition_Tor),
-					Blades: make(map[int64]*pb.Definition_Blade),
+					Pdus:   make(map[int64]*pb.Definition_Pdu, len(xfrRack.Pdus)),
+					Tors:   make(map[int64]*pb.Definition_Tor, len(xfrRack.Tors)),
+					Blades: make(map[int64]*pb.Definition_Blade, len(xfrRack.Blades)),
 				}
 
 				// A rack contains sub-components for Pdus, Tors and Blades with
@@ -699,7 +699,7 @@ func toDefinitionRoot(xfrRoot *rootEx) (*pb.Definition_Root, error) {
 							Condition: condition,
 							Enabled:   xfrPdu.Details.Enabled,
 						},
-						Ports: make(map[int64]*pb.PowerPort),
+						Ports: make(map[int64]*pb.PowerPort, len(xfrPdu.Ports)),
 					}
 
 					// Iterate over the set of power ports in the supplied
@@ -779,7 +779,7 @@ func toDefinitionRoot(xfrRoot *rootEx) (*pb.Definition_Root, error) {
 							Condition: condition,
 							Enabled:   xfrTor.Details.Enabled,
 						},
-						Ports: make(map[int64]*pb.NetworkPort),
+						Ports: make(map[int64]*pb.NetworkPort, len(xfrTor.Ports)),
 					}
 
 					// Iterate over the set of network ports in the supplied
@@ -906,9 +906,9 @@ func toDefinitionRoot(xfrRoot *rootEx) (*pb.Definition_Root, error) {
 			return nil, errors.ErrRegionValidationFailure{
 				Region: xfrRegion.Name,
 				Err:  err,
+			}
 		}
 	}
-}
 
 	return root, nil
 }
--- a/simulation/internal/clients/inventory/reader.go
+++ b/simulation/internal/clients/inventory/reader.go
@@ -5,6 +5,7 @@ package inventory
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/spf13/viper"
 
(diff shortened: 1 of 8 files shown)


simulation/internal/clients/inventory/reader.go
```diff
@@ -334,3 +447,468 @@ func toDefinitionRegionInternal(xfr *zone) (*pb.Definition_Region, error) {
...
+	// and a set of blades.
+	//
+	for _, xfrRegion := range xfrRoot.Regions {
+		if _, ok := root.Regions[xfrRegion.Name]; ok {
+			return nil, errors.ErrConfigRegionDuplicate{
+				Region: xfrRegion.Name,
+			}
+		}
+		var (
+			ok		   bool
```
Jim3Things, line 576: Weird alignment

simulation/internal/clients/inventory/reader.go
```diff
@@ -334,3 +447,468 @@ func toDefinitionRegionInternal(xfr *zone) (*pb.Definition_Region, error) {
...
+			}
+
+			zone := &pb.Definition_Zone{
+				Details: &pb.ZoneDetails{
+					State:    state,
+					Enabled:  xfrZone.Details.Enabled,
+					Location: xfrZone.Details.Location,
+					Notes:    xfrZone.Details.Notes,
+				},
+				Racks: make(map[string]*pb.Definition_Rack),
```
Jim3Things, line 629: I've been puzzling over whether to include the allocation amount in these make calls.  Racks and blades are likely to be the only ones where there are a lot of entries, so I ended up asking the question here.  Up to you, just a question.

simulation/internal/clients/inventory/reader.go
```diff
@@ -334,3 +447,468 @@ func toDefinitionRegionInternal(xfr *zone) (*pb.Definition_Region, error) {
...
+	return hwType, ok
+}
+
+// toDefinitionRoot converts intermediate values to the final format
+// One important difference is that the intermediate is array based.
+// The final format is map based using specific fields in array
+// entries as the map keys
+//
+func toDefinitionRoot(xfrRoot *rootEx) (*pb.Definition_Root, error) {
+
```
Jim3Things, line 556: Stylistically, given how long this function is, I'd look to split it up with smaller targeted functions (e.g. one to handle blades in a rack, racks in a zone, zones in a region, and so on).

simulation/internal/clients/inventory/reader.go
```diff
@@ -334,3 +447,468 @@ func toDefinitionRegionInternal(xfr *zone) (*pb.Definition_Region, error) {
...
+var methodMap = map[string]pb.BladeBootInfo_Method{
+	"local": pb.BladeBootInfo_local,
+	"network": pb.BladeBootInfo_network,
+}
+
+func getBootMethodFromString(s string) (pb.BladeBootInfo_Method, bool) {
+	method, ok := methodMap[strings.ToLower(s)]
+
+	return method, ok
+}
```
Jim3Things, line 535: I keep thinking that these can be made simpler in their use.  Every call turns the boolean into an error, but the error takes the hierarchy to fill in the error text.  I keep thinking that there must be a way with the types and methods on that type that these could be supplanted by something that does that automatically.  Not a solution to present just yet, unless these are always methods on the root and are passed the child address in the object tree?

simulation/internal/clients/inventory/reader_test.go
```diff
@@ -250,10 +250,215 @@ func (ts *readerTestSuite) TestReadInventoryDefinitionFromFileValidateBlade() {
 	require := ts.Require()
 
 	response, err := ReadInventoryDefinitionFromFile(context.Background(), "./testdata/BadYamlValidate")
-	require.EqualError(err, "In rack \"rack1\": the field \"Blades[2].Cores\" must be greater than or equal to 1.  It is 0, which is invalid")
+	require.EqualError(err, "In rack \"rack1\": the field \"Blade[2].Cores\" must be greater than or equal to 1.  It is 0, which is invalid")
```
Jim3Things, line 253: Why did the collection field name go from plural to singular?

simulation/internal/clients/inventory/reader_test.go
```diff
@@ -250,10 +250,215 @@ func (ts *readerTestSuite) TestReadInventoryDefinitionFromFileValidateBlade() {
...
+	// s, ok := response.Racks["rack2"]
+	// require.True(ok)
+	// assert.Equal(2, len(s.Blades))
+
+	// c, ok := r.Blades[2]
+	// require.True(ok)
+	// assert.Equal(int64(8), c.Cores)
+	// assert.Equal(int64(16834), c.MemoryInMb)
+	// assert.Equal(int64(120), c.DiskInGb)
+	// assert.Equal(int64(2048), c.NetworkBandwidthInMbps)
```
Jim3Things, line 459: If these are now stale, remove.

simulation/internal/clients/inventory/testdata/Extended/inventory.yaml
```diff
@@ -0,0 +1,116 @@
...
+              Location: "Pacific NW, row 1, rack 1"
+              Notes: "Simple rack definition"
+            Pdus:
+              - Index: 0
+                Details:
+                  Enabled: true
+                  Condition: Operational
+                Ports:
+                  - Index: 1
+                    Wired: 1
```
Jim3Things, line 31: Shouldn't Wired be true or false?

simulation/pkg/protos/inventory/definition.Validate.go
```diff
@@ -275,7 +314,7 @@ func (x *Definition_Rack) Validate(prefix string) error {
 	// .. And then validate that each blade is valid
 	//
 	for k, v := range x.Blades {
-		if err := v.Capacity.Validate(fmt.Sprintf("%sBlades[%d].", prefix, k)); err != nil {
+		if err := v.Validate(fmt.Sprintf("%sBlade[%d].", prefix, k)); err != nil {
```
Jim3Things, line 317: Why does this not stay as Blades?

simulation/internal/clients/inventory/reader.go
MikeCJohnsonUk: How so? There were some embedded tabs, so I've remove those and used spaces instead. Would that answer it?

simulation/internal/clients/inventory/reader.go
MikeCJohnsonUk: Well, my usual rule of thumb is that if the quantity is know, then use in in the construction. The difficulty here is that while we do know the number of candidates for rack entries, they have not yet been tested for duplicates.

So I'm in a quandary: we know the upper bound, but have only limited confidence in the actual number. If we assume that the majority of inventory definitions are likely correct, then we can use the candidate number. If not...

I think I'm going to choose to use the upper bound number. Worst case is that we allocate a little extra memory for a while.

Make sense?

simulation/internal/clients/inventory/reader.go
MikeCJohnsonUk: Yeah, I pretty much the same notion as I got close to completing it. I'll take a look.

simulation/pkg/protos/inventory/definition.Validate.go
MikeCJohnsonUk: Well, we have singular for Tor and Pdu, so for consistency. It would probably be slightly better to go with all of them being plural. I'll change them.

simulation/internal/clients/inventory/testdata/Extended/inventory.yaml
MikeCJohnsonUk: Yes. Enabled has the same issue.

Fixed

simulation/internal/clients/inventory/reader_test.go
MikeCJohnsonUk: Removed.

simulation/internal/clients/inventory/reader.go
MikeCJohnsonUk: The only reason there is a getXxxFromString() function at all is to get over the inability to declare the map as a const. At least the functions are about as simple as possible and single purpose.

I guess you could change them to take the object, or the path elements so that they could generate the error themselves, but that seems to be adding a bunch of stuff just to move the error generation from the call site to inside the function. Not sure that's a good tradeoff.

One way that would work is for the call site to construct the error and then pass that in to the function to be returned as required, but that now requires that the error be built regardless of whether or not it's needed.

simulation/internal/clients/inventory/reader.go
MikeCJohnsonUk: Took a look. It runs into exactly the same problem, related actually, as the passing in of all the current context to obtain the "path" to the object of interest.

I think I have a way forwards here, and it is associated with how to get a generalized address mechanism to work. Unfortunately it's quite a chunk of code and really to much to do here. As a consequence, I think this should be deferred until that mechanism is available. I've added issue #221

simulation/internal/clients/inventory/reader.go
MikeCJohnsonUk: See issue #221

simulation/internal/clients/inventory/reader.go
MikeCJohnsonUk: Added allocation.

simulation/internal/clients/inventory/reader_test.go
MikeCJohnsonUk: See below.

simulation/internal/clients/inventory/reader.go
Jim3Things: Probably - it was because on the github diff display the bool was way off from the pb.State on line 577.  Embedded tabs could cause that.

simulation/pkg/protos/inventory/definition.Validate.go
Jim3Things: I'd prefer that we use plural for maps.  I know that Tor and Pdu started as singular, so making them Tors and Pdus makes sense to me.

simulation/internal/clients/inventory/testdata/Extended/inventory.yaml
Jim3Things: ok

simulation/internal/clients/inventory/reader.go
Jim3Things: ok

simulation/internal/clients/inventory/reader.go
Jim3Things: ok

simulation/pkg/protos/inventory/definition.Validate.go
MikeCJohnsonUk: Ok, I'll update them all to plural.

Jim3Things (approved):

PR #225, opened 2021-02-21
Update testify version

Jim3Things:


MikeCJohnsonUk (approved):
I'll admit to not looking too closely at the changes since they're under vendor, but some at least looked pretty interesting.

PR #227, opened 2021-02-22
Move session management constants to cloudchamber.yaml

Jim3Things:
The constants that determine the inactivity time and maximum simultaneous session limit are moved to the yaml file.

This also prompted fully encapsulating the session manager into a struct/interface, and having the instance be part of the frontend server instance.

### DIFF
--- a/simulation/internal/config/settings.go
+++ b/simulation/internal/config/settings.go
@@ -34,11 +34,11 @@ const (
 	simSupportDefaultPort      = 8083
 	simSupportDefaultTraceFile = ".\\sim_support_trace.txt"
 
-	webServerDefaultPort      = 8084
-	webServerFEDefaultPort    = 8080
-	webServerDefaultTraceFile = ".\\web_server_trace.txt"
-	defaultHost               = ""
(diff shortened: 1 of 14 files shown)


simulation/internal/config/settings.go
```diff
@@ -194,6 +199,8 @@ func newGlobalConfig() *GlobalConfig {
 			RootFilePath:          defaultRootFilePath,
 			SystemAccount:         defaultSystemAccount,
 			SystemAccountPassword: defaultSystemPassword,
+			SessionInactivity: webServerDefaultInactivity,
+			ActiveSessionLimit: webServerDefaultSessionLimit,
```
MikeCJohnsonUk, line 203: Alignment?

MikeCJohnsonUk (commented):
I feel so picky with just a *potential* for an alignment issue as the only thing.

simulation/internal/config/settings.go
Jim3Things: fixed.

MikeCJohnsonUk (approved):

PR #233, opened 2021-02-28
Add missing region manipulation routines (closes #229)

### LINKED ISSUE #229: Add ability to DBInventory to manipulate regions
The DBInventory layer currently is missing some of the region object manipulation APIs, such as CreateRegion(), ReadRegion(), DeleteRegion() etc.

These need to be added along with the appropriate tests.

MikeCJohnsonUk:
Add routines with associated tests to manipulation regions from the DBInventory layer.

### DIFF
--- a/simulation/internal/services/frontend/DBInventory.go
+++ b/simulation/internal/services/frontend/DBInventory.go
@@ -547,6 +547,59 @@ func WithRevision(rev int64) InventoryZoneOption {
 	return func(options *InventoryZoneOptions) { options.revision = rev }
 }
 
+// Region options
+
+// InventoryRegionOption is a
+//
+type InventoryRegionOption func(*InventoryRegionOptions)
+
+// InventoryRegionOptions is a struct
+//
+type InventoryRegionOptions struct {
+	revision      int64
+	includeZones  bool
+	includeRacks  bool
+	includePdus   bool
+	includeTors   bool
+	includeBlades bool
+}
+
+func (options *InventoryRegionOptions) applyRegionOpts(optionsArray []InventoryRegionOption) {
+	for _, option := range optionsArray {
+		option(options)
+	}
+}
+
+// WithRegionRevision is a
+//
+func WithRegionRevision(rev int64) InventoryRegionOption {
+	return func(options *InventoryRegionOptions) { options.revision = rev }
+}
+
+// WithRegionRacks is a
+//
+func WithRegionRacks() InventoryRegionOption {
+	return func(options *InventoryRegionOptions) { options.includeRacks = true }
+}
+
+// WithRegionTors is a
+//
+func WithRegionTors() InventoryRegionOption {
+	return func(options *InventoryRegionOptions) { options.includeTors = true }
+}
+
+// WithRegionPdus is a
+//
+func WithRegionPdus() InventoryRegionOption {
+	return func(options *InventoryRegionOptions) { options.includePdus = true }
+}
+
+// WithRegionBlades is a
+//
+func WithRegionBlades() InventoryRegionOption {
+	return func(options *InventoryRegionOptions) { options.includeBlades = true }
+}
+
 // Zone options
 
 // InventoryZoneOption is a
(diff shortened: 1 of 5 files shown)


simulation/internal/services/frontend/DBInventory.go
```diff
@@ -1072,6 +1184,37 @@ func (m *DBInventory) ReadBlade(
...
+//
+func (m *DBInventory) UpdateRegion(
+	ctx context.Context,
+	name string,
+	region *pb.Definition_Region,
+	options ...InventoryRegionOption) (int64, error) {
+
+	z, err := m.inventory.NewRegion(
+		inventory.DefinitionTable,
+		name)
```
MikeCJohnsonUk, line 1201: Minor typo - variable name for created region should be 'r'.

Multiple instance in function.

simulation/internal/services/frontend/DBInventory.go
```diff
@@ -710,6 +763,37 @@ func (m *DBInventory) ListBlades(
...
+//
+// This record created will contain just the region level details
+// and any additional zone, rack, blade, tor or pdu data will be ignored
+// and not included in the stored record.
+//
+func (m *DBInventory) CreateRegion(
+	ctx context.Context,
+	name string,
+	region *pb.Definition_Region,
+	options ...InventoryRegionOption) (int64, error) {
```
Jim3Things, line 776: It doesn't look like the options are used -?

I'm not sure what to comment on the InventoryRegionOption instances, or the various options parameters.  It doesn't look like they're used -- is that because this is part of a change that is coming later, have I missed the references, or something else?

MikeCJohnsonUk: At present, not used. I Will need to get the WithRevision() option at least to work though I'm somewhat less certain about the others. That would actually help quite a bit as that's where a lot of the extra stuff comes from.

I just added the InventoryRegionOptions for now to echo the structure of the other calls.

Jim3Things: ok.

simulation/internal/services/frontend/DBInventory_test.go
```diff
@@ -644,6 +641,18 @@ func (ts *DBInventoryTestSuite) equalNetworkPort(expected *pb.NetworkPort, actua
...
+
+	if !assert.Equal(expected.State, actual.State) ||
+	!assert.Equal(expected.Location, actual.Location) ||
+	!assert.Equal(expected.Notes, actual.Notes) {
+		return false
+	}
+
+	return true
+}
+
```
Jim3Things, line 655: consider assert/require.Condition, and make this a function that can be used as the compare parameter...

MikeCJohnsonUk: Yes, but wouldn't that lose the indication on which field triggered the assertion? If the expectation is that which of the individual field failures is not significant, which may well be true here, then the individual comparisons could all be collapsed to a single bit and then assert on the bit, which is essentially the same as using .Condition() with a comparison function. I'm not sure I see the benefit.

Jim3Things: It was the structure of this which caused my comment.  I've personally been leaning increasingly to require.Xxx, rather than assert.Xxx - and if you were using require.Xxx, then this function would be literally three require.Equal statements with a void return.  It was because it did a bunch of tests to return a bit that caused my comment.

Willing to go with any of the options, frankly.

MikeCJohnsonUk: Well, sorting this out proved to be quite the adventure. Underlying problem which lead to this was the protobuf internal field state.atomicMessageInfo which is set as a result of an JSON encode or decode operation on the underlying fields. Since structs often had references which originated in earlier structs operating on the "source" struct could cause the change in the reference child struct which might also be referenced elsewhere. 

Since the testify require / assert Equal() function does not ignore the state.atomicMessageInfo field, tests would fail for non-obvious reasons which lead to the original field-by-field asserts and then onto the tiered verification functions.

Instead, the product code has been altered to isolate child structs and generate new child structs rather than merely using references to the original child structs. Although not ideal, this works and is safe.

A better approach, and a suggestion for a long term solution is to extend testify and teach it how to properly compare protobuf generated objects.

Jim3Things (approved):

PR #234, opened 2021-03-03
Active state persistence, part 1

Jim3Things:
This adds the save-to-a-protobuf part of the state machine. Restore, and actual connection to the store are later.

Rack and stepper are mostly latent support to be filled in later.

### DIFF
--- a/simulation/internal/services/inventory/blade.go
+++ b/simulation/internal/services/inventory/blade.go
@@ -2,7 +2,6 @@ package inventory
 
 import (
 	"context"
-	"fmt"
 	"math/rand"
 
 	"github.com/golang/protobuf/proto"
@@ -264,7 +263,7 @@ func bootingTimerExpiry(ctx context.Context, machine *sm.SM, m sm.Envelope) bool
 
 // bootingOnLeave ensures that any active boot delay timer is canceled before
 // proceeding to a non-booting state.
-func bootingOnLeave(ctx context.Context, machine *sm.SM, nextState fmt.Stringer) {
+func bootingOnLeave(ctx context.Context, machine *sm.SM, nextState sm.StateIndex) {
 	if nextState != pb.Actual_Blade_booting {
 		cancelTimer(ctx, machine, "boot")
 	}
@@ -279,7 +278,7 @@ func stoppingTimerExpiry(ctx context.Context, machine *sm.SM, m sm.Envelope) boo
 
 // stoppingOnLeave ensures that any active time is canceled before proceeding
 // to a non-stopping state.
-func stoppingOnLeave(ctx context.Context, machine *sm.SM, nextState fmt.Stringer) {
+func stoppingOnLeave(ctx context.Context, machine *sm.SM, nextState sm.StateIndex) {
 	if nextState != pb.Actual_Blade_stopping &&
 		nextState != pb.Actual_Blade_stopping_isolated {
 		cancelTimer(ctx, machine, "shutdown")
--- a/simulation/internal/services/inventory/blade.go
+++ b/simulation/internal/services/inventory/blade.go
@@ -2,7 +2,10 @@ package inventory
 
 import (
 	"context"
+	"fmt"
 	"math/rand"
+
+	"github.com/golang/protobuf/proto"
 
 	"github.com/Jim3Things/CloudChamber/simulation/internal/common"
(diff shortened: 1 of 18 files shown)


simulation/internal/services/inventory/blade.go
```diff
@@ -291,8 +264,8 @@ func bootingTimerExpiry(ctx context.Context, machine *sm.SM, m sm.Envelope) bool
 
 // bootingOnLeave ensures that any active boot delay timer is canceled before
 // proceeding to a non-booting state.
-func bootingOnLeave(ctx context.Context, machine *sm.SM, nextState string) {
-	if nextState != bladeBooting {
+func bootingOnLeave(ctx context.Context, machine *sm.SM, nextState fmt.Stringer) {
+	if nextState != pb.Actual_Blade_booting {
```
MikeCJohnsonUk, line 268: Ah, surely the type of nextState is a member of the set of actual blade State enum, that is the type should be a pb.Actual_Blade_State, or whatever it resolves out to be rather than some thing that happens to implement the Stringer interface?

I've noticed a number of places where the this seems to be happening, so I could well be missing something here. I can see why there needs to be a String() method for the type (i.e. the type implements the fmt.Stringer interface), but that doesn't alter the type itself. Right?

simulation/internal/services/inventory/pdu.go
```diff
@@ -127,7 +139,7 @@ func pduGetStatus(ctx context.Context, machine *sm.SM, m sm.Envelope) bool {
 
 	pduStatus := &messages.PduStatus{
 		StatusBody: messages.StatusBody{
-			State:     p.sm.CurrentIndex,
+			State:     p.sm.CurrentIndex.String(),
```
MikeCJohnsonUk, line 142: Ah, as above re: the type vs an interface implemented by the type?

simulation/internal/services/stepper/stepper.go
```diff
@@ -600,15 +579,15 @@ func checkForExpiry(ctx context.Context, machine *sm.SM) {
 
 // convertFromState translates the current state into a policy enum value that
 // can be used in the GetStatus response message.
-func convertFromState(state string) int {
+func convertFromState(state fmt.Stringer) int {
```
MikeCJohnsonUk, line 582: As above re: the type vs an interface implemented by the type?

simulation/internal/sm/actionState.go
```diff
@@ -2,23 +2,22 @@ package sm
...
 var (
 	// NullEnter defines a state entry function that does nothing.
 	NullEnter EnterFunc = nil
 
 	// NullLeave defines a state exit function that does nothing.
 	NullLeave LeaveFunc = nil
+
+	// Stay is used in an ActionEntry as shorthand to indicate that there is
+	// no state transition to process.
+	Stay fmt.Stringer = nil
```
MikeCJohnsonUk, line 20: As above re: the type vs an interface implemented by the type?

simulation/internal/services/inventory/blade.go
Jim3Things: Yeah, the problem that I ran into is that the XxxLeave functions need to be implementations of LeaveFunc, which is defined independently of any particular state machine (LeaveFunc is the signature that the core state machine uses when leaving).  Now, that really can only know some general interface type, and fmt.Stringer was the most specific I could see.  If I make bootingOnLeave use the pb.Actual_Blade_State the reference to it gets a failure, even though you'd think it was compatible. So, I made sure that the signature matched the LeaveFunc definition exactly...

simulation/internal/services/inventory/pdu.go
Jim3Things: ?  StatusBody uses a string here, is the suggestion to use the enum value?

simulation/internal/services/stepper/stepper.go
Jim3Things: Yeah, this is the same basic problem as the first case.

simulation/internal/sm/actionState.go
Jim3Things: Not sure I'm following here.

simulation/internal/services/inventory/blade.go
Jim3Things: talked offline, and think we have an intermediate approach.

MikeCJohnsonUk (approved):

PR #236, opened 2021-03-04
Unit test structure improvements

Jim3Things:
Move trailing UTs to testify suite
Simplify timestamp_test logic.

### DIFF
--- a/simulation/internal/clients/timestamp/timestamp_test.go
+++ b/simulation/internal/clients/timestamp/timestamp_test.go
@@ -17,28 +17,26 @@ type timestampTestSuite struct {
 	stepperTestClientSuite
 }
 
+func (ts *timestampTestSuite) verifyNow(ctx context.Context, tick int64) {
+	require := ts.Require()
+
+	now, err := Now(ctx)
+	require.NoError(err)
+	require.Equal(tick, now.Ticks)
+}
+
 func (ts *timestampTestSuite) TestNow() {
 	require := ts.Require()
-	assert := ts.Assert()
 
 	ctx := context.Background()
 
-	now, err := Now(ctx)
-	require.Nil(err)
-	assert.Equal(int64(0), now.Ticks)
+	ts.verifyNow(ctx, 0)
+	ts.verifyNow(ctx, 0)
 
-	now, err = Now(ctx)
-	require.Nil(err)
-	assert.Equal(int64(0), now.Ticks)
+	require.NoError(Advance(ctx))
 
-	assert.Nil(Advance(ctx))
-	now, err = Now(ctx)
-	require.Nil(err)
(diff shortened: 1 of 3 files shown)


simulation/internal/clients/timestamp/timestamp_test.go
```diff
@@ -47,28 +45,24 @@ func (ts *timestampTestSuite) TestTimestamp_After() {
...
+	require.NoError(Advance(ctx))
+	require.NoError(Advance(ctx))
+	assert.True(common.DoNotCompleteWithin(ch, 2*time.Second))
 
-	require.Nil(Advance(ctx))
-	assert.True(common.DoNotCompleteWithin(ch, time.Duration(2)*time.Second))
-
-	require.Nil(Advance(ctx))
-	assert.True(common.CompleteWithin(ch, time.Duration(2)*time.Second))
+	require.NoError(Advance(ctx))
```
MikeCJohnsonUk, line 64: We've lost a call to DoNotCompleteWithin() - deliberate?

simulation/internal/common/decode_test.go
```diff
@@ -1,102 +1,104 @@
...
+}
 
-	got, err := Decode(emptyArray, someString)
-	require.Error(t, err)
-	assert.Equal(t, expected, got)
+func (ts *decodeTestSuite) TestDecodeBlankSource() {
+	ts.doFail(
+		ts.keywords(),
+		"",
+		`The source "" was not found in keywords [Defined Actual Target Observed].`)
```
MikeCJohnsonUk, line 64: Ah, a little unusual print format string to indicate an empty source/test string?

simulation/internal/clients/timestamp/timestamp_test.go
Jim3Things: Yeah - I put it in early to help with debugging, just in case.  But semantically all it did was to cause the test to run 2s longer than it needed to...

simulation/internal/common/decode_test.go
Jim3Things: I'm guessing this has to do with the back-carat?  Yeah, that lets me use " without escaping it.

simulation/internal/clients/timestamp/timestamp_test.go
MikeCJohnsonUk: Ok

simulation/internal/common/decode_test.go
MikeCJohnsonUk: It was the double double-quote I was referring to rather than the back-carat. I mean, it's not wrong, I just thought it might look a little odd printing out double double-quote to indicate an empty string to the user.

MikeCJohnsonUk (approved):

PR #242, opened 2021-03-07
Replace all uses of earlier inventory yaml format with extended version (Closes #228, closes #235)

### LINKED ISSUE #228: Update remaining uses of the older inventory schema the newer enhanced variant
Some of the tests in inventory\reader and the code itself are still using the earlier variant of the inventory yaml schema. This need to be updated to use the newer variant.

### LINKED ISSUE #235: Update all services to make use of extended inventory schema

MikeCJohnsonUk:
Updates the remaining existing inventory.yaml files to use the extended format
Removes the routines to read and process the earlier format
Removes the tests for the routines processing the earlier format
Updates the inventory service to load and process the extended inventory file
Updates the web server and inventory services so that the inventory service is the only service to load the configured inventory into the store

### DIFF
--- a/simulation/internal/clients/inventory/reader.go
+++ b/simulation/internal/clients/inventory/reader.go
@@ -22,45 +22,6 @@ const (
 
 // +++ Intermediate binary format
 
-// This mirrors the external zone format except the keys
-// are fields of array
-
-type zone struct {
-	Racks []rack
-}
-
-type rack struct {
-	Name   string
-	Blades []blade
-	Tor    tor
-	Pdu    pdu
-}
-
-type blade struct {
-	Index                  int64
-	Arch                   string
-	Cores                  int64
-	DiskInGb               int64
-	MemoryInMb             int64
-	NetworkBandwidthInMbps int64
-}
-
-type tor struct {
-}
-
-type pdu struct {
-}
-
-type rootEx struct {
-	ZoneTypes   []catZone
-	RackTypes   []catRack
-	PduTypes    []catPdu
-	TorTypes    []catTor
-	BladeTypes  []catBlade
-	Details     rootExDetails
-	Regions     []regionEx
-}
-
 // ++++++++++
 //
 // The folowing catXxx structs are used as throwaway data fields to enable
@@ -112,6 +73,21 @@ type catPort struct {
 
 // ----------
 
+
+// This struct is used to load the inventory definition file in a
+// temporary format prior to converting to something the rest of
+// the system understand and using that to update the persisted
+//inventory in the store.
+//
+type rootEx struct {
+	ZoneTypes   []catZone
+	RackTypes   []catRack
+	PduTypes    []catPdu
+	TorTypes    []catTor
+	BladeTypes  []catBlade
+	Details     rootExDetails
+	Regions     []regionEx
+}
 
 type rootExDetails struct {
 	Name  string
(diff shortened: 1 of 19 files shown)


simulation/internal/clients/inventory/reader_test.go
```diff
@@ -41,82 +41,63 @@ func (ts *readerTestSuite) TestReadInventoryDefinition() {
...
-	require := ts.Require()
-
-	response, err := ReadInventoryDefinition(context.Background(), "./testdata/BadYamlValidate")
-	require.EqualError(err, "In rack \"rack1\": the field \"Blades[2].Cores\" must be greater than or equal to 1.  It is 0, which is invalid")
-	require.Nil(response)
+	require.NotNil(b.Capacity)
+	assert.Equal(int64(8), c.Capacity.Cores)
+	assert.Equal(int64(16384), c.Capacity.MemoryInMb)
+	assert.Equal(int64(120), c.Capacity.DiskInGb)
+	assert.Equal(int64(2048), c.Capacity.NetworkBandwidthInMbps)
```
Jim3Things, line 80: This is what we've been using, and I'm good with that.  They're examples where we could use EqualValues and avoid the typecast with the constants.  Not sure it is worth it.  So just fyi.

simulation/internal/clients/inventory/reader_test.go
```diff
@@ -41,82 +41,63 @@ func (ts *readerTestSuite) TestReadInventoryDefinition() {
 	assert := ts.Assert()
 	require := ts.Require()
 
-	response, err := ReadInventoryDefinition(context.Background(), "./testdata/Simple")
+	response, err := ReadInventoryDefinitionFromFileEx(context.Background(), "./testdata/Simple")
```
Jim3Things, line 44: Will this be renamed to drop the Ex?

simulation/internal/clients/inventory/testdata/BadYaml/inventory.yaml
```diff
@@ -1,34 +1,85 @@
-#defining a blade1 and blade2 for rack1
-Racks:
-  - Name: rack1
-    Blades:
-    - Index: 1
+# Simple test configuration comprising 1 region, with 1 zone, with the zone having 2 racks, with each rack having a single pdu, a single tor and 2 blades.
+#
+# Two racks, with a duplicated rack name.
+
+BladeTypeS:
```
Jim3Things, line 5: BladeTypeS, not BladeTypes?

simulation/internal/clients/inventory/testdata/BadYaml/inventory.yaml
```diff
@@ -1,34 +1,85 @@
-#defining a blade1 and blade2 for rack1
-Racks:
-  - Name: rack1
-    Blades:
-    - Index: 1
+# Simple test configuration comprising 1 region, with 1 zone, with the zone having 2 racks, with each rack having a single pdu, a single tor and 2 blades.
```
Jim3Things, line 1: I don't see the reference yaml here - did I miss it?

Regardless, it would be good to have a comment in the yaml files that we provide that lists the variables and what they are replaced with.   I can guess on $NODENAME$, but, well, it's a guess...

simulation/internal/clients/inventory/reader.go
```diff
@@ -552,11 +246,11 @@ func ReadInventoryDefinitionFromFileEx(ctx context.Context, path string) (*pb.De
...
-    "burn_in":        pb.Condition_burn_in,
-    "out_for_repair": pb.Condition_out_for_repair,
-    "retiring":       pb.Condition_retiring,
-    "retired":        pb.Condition_retired,
+	"operational":    pb.Condition_operational,
+	"burn_in":        pb.Condition_burn_in,
+	"out_for_repair": pb.Condition_out_for_repair,
+	"retiring":       pb.Condition_retiring,
+	"retired":        pb.Condition_retired,
 }
```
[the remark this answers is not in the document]
MikeCJohnsonUk: So there is. Minor complication in that the generated func returns an int32 rather than the enum, but a quick cast sorts that out.

simulation/internal/clients/inventory/reader_test.go
MikeCJohnsonUk: I've given it a try. Now I just need to remember to use it elsewhere too.

simulation/internal/clients/inventory/reader_test.go
MikeCJohnsonUk: That was my intention...

Done

simulation/internal/clients/inventory/testdata/BadYaml/inventory.yaml
MikeCJohnsonUk: Yup. Though strictly speaking, the actual string here is not important, as long as something is there.

Done.

simulation/internal/clients/inventory/testdata/BadYaml/inventory.yaml
MikeCJohnsonUk: The reference yaml was in a previous PR, already in the extended format. No changes in this PR.

As for the $NODENAME$ string, this was just me experimenting to check what would make it into the internal variables and currently has no significance. I expect that something like this will be in a future PR along the lines of what we discussed, but no plans on that to date. At that point, having a convention to comment on the variables would be a good move.

Jim3Things (approved):

PR #249, opened 2021-03-15
Add ability to establish Watchpoints on portion of the store namespace, part 1

MikeCJohnsonUk:
Add feature to the store to allow a client to establish one or more watchpoint on portions of the store namespace to detect changes to inventory, workloads, target, observed states etc.

This change provides the underlying primitives at the lower layer of the store which will establish watchpoints and translate from the etcd layer eventing mechanism to something a little more convenient.

This will be followed by part 2 which will contain the storeapi layer to connect the store primitives provided in this change to the store clients using concepts those clients can process.

As a result, changes to storeapi.go can be ignored for the present if desired, although part 2 will contain changes resembling those here but at a later date.

Also includes some minor changes related to passing in an external context for enhanced tracing in the store initialization paths.

### DIFF
--- a/simulation/internal/clients/store/store.go
+++ b/simulation/internal/clients/store/store.go
@@ -66,7 +66,6 @@ package store
 import (
 	"context"
 	"fmt"
-	"log"
 	"strings"
 	"sync"
 	"time"
@@ -202,9 +201,11 @@ func (store *Store) disconnected(ctx context.Context) error {
 // Initialize is a method used to initialise the basic global state used to access
 // the back-end db service.
 //
-func Initialize(cfg *config.GlobalConfig) {
-	ctx, span := tracing.StartSpan(context.Background(),
-		tracing.WithContextValue(timestamp.OutsideTime))
+func Initialize(ctx context.Context, cfg *config.GlobalConfig) {
+	ctx, span := tracing.StartSpan(ctx,
+		tracing.WithContextValue(timestamp.OutsideTime),
+		tracing.AsInternal(),
+		tracing.WithName("Initialize store"))
 	defer span.End()
 
 	storeRoot.DefaultEndpoints = []string{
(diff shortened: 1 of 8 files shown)


simulation/internal/clients/store/store.go
```diff
@@ -696,39 +698,178 @@ func (store *Store) UpdateClusterConnections() error {
...
+	WatchEventTypeCreate WatchEventType = iota
+
+	// WatchEventTypeUpdate indicates the event was an Update or modify operation
+	//
+	WatchEventTypeUpdate
+
+	// WatchEventTypeDelete indicates the event was a Delete operation
+	//
+	WatchEventTypeDelete
+)
```
Jim3Things, line 728: Question only: can the store end up finding that it has missed some changes?  If so, what does that look like for someone who has set up a watch listener?

Jim3Things (approved):

simulation/internal/clients/store/store.go
MikeCJohnsonUk: Not sure I understand the question. The store keeps no real state as such, so it has no notion of missing changes. It's conceivable that a change happed though something other than the store layer, e.g. via the etcdctl utility direct to etcd, in which case the Watch event should trigger if the change was made on a monitored key/portion of the namespace.

I get the feeling I'm missing the significance of the question. Could you expand on that? Missing changes from whos perspective?

Jim3Things: My question was pretty vague because I'm not sure how etcd watch works exactly.  It was prompted by wondering if there were any cases where updates to etcd could overrun the ability to notify watchers of changes, and if so, in what way.  I'm not sure a) that it is possible, or b) we care about outdated changes to the same data item (i.e. the watcher only reports on changes 1 & 3 to an element, because 3 happened before the change notification for 2 could be posted).  But I am sure that a lost change notification between two elements would matter (i.e. a change to item 'a' is not reported because a change to item 'b' somehow prevented that).

So I ended up just asking - whether there were cases where a watch could end up not reporting on a data change, and if so, in what way, and if it mattered to the listener, how they could detect or account for it.

MikeCJohnsonUk: There are some guarantees about Watch events (see https://etcd.io/docs/v3.4.0/learning/api/#watch-api) but the client we use has wrapped the underlying calls and then we wrap that. As far as I can tell, we do nothing to void those guarantees and since we do capture/forward the data and revision from the previous state, it should be possible to tell if one or more events has been missed or overlooked. We also report the current store revision so it should be possible to determine if global ordering is somehow broken too.

I *think* that answers your question.

The only question I have, and I've not seen it answered, is how deep is the pipeline? That is, what happens if the events are not "consumed"?

MikeCJohnsonUk: Update: I've just noticed something that can be construed to mean that Watch events are generated by running a cursor along the *log* for the system. By default, Watch events are started with whatever "now" is when initiated, but you can start from anywhere in history providing that history has not been compacted away.

If my interpretation is accurate, that would imply that failing to consume events just results in failing to update the pointer in history, and would just result in the sequence of Watch events getting further and further behind "now". That is watch events are not guaranteed to be real-time, and also that failing to consume an event will not interrupt the main db update stream.

Currently, we do not set the starting revision, but there is an option for that we can employ should we wish.

PR #270, opened 2021-04-17
Use the grpc internal backoff mechanisms for connection retry

### LINKED ISSUE #259: Use GRPC's supported connection options to drive retries on failure
Remove the manual backoff logic and use the GRPC included options.

Jim3Things:
Also fixed two small issues that I found when testing stop/restart of sim_supportd while the UI was active.

closes #259


simulation/internal/services/frontend/frontend.go
```diff
@@ -25,6 +25,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/Jim3Things/CloudChamber/simulation/internal/clients/limits"
```
manpreet85, line 28: Is it in the correct place?

Jim3Things: Nope, it's out of order. I've fixed this and the other places it happened.

simulation/internal/services/frontend/logs.go
```diff
@@ -60,7 +60,7 @@ func handlerLogsGetAfter(w http.ResponseWriter, r *http.Request) {
 
 	data := <-ch
 	if data.Err != nil {
-		postHTTPError(ctx, w, err)
+		postHTTPError(ctx, w, data.Err)
 		return
 	}
```
MikeCJohnsonUk, line 65: Can the channel have been closed *before* anything was written to it? That is, can this read from the channel fail, i.e. is "data" garunteed non-nil?

Jim3Things: Yes, the channel should have one message in it before it is closed.  I didn't check for nil under the theory that if the internal consistency fails in this way, I'd rather have a panic...

simulation/internal/services/stepper/stepper_test.go
```diff
@@ -7,6 +7,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/Jim3Things/CloudChamber/simulation/internal/clients/limits"
```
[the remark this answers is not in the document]
Jim3Things: This should have already been done - old version?

MikeCJohnsonUk (approved):

PR #273, opened 2021-04-18
Update to latest react and dependency versions

Jim3Things:
Note that we do not check in the node_manager library - it is fetched from package.json.

After making these changes, I did:
npm install (which installs the dependencies from package.json)
npm prune (which did nothing on my system, fwiw)
npm find-dupes --legacy-peer-deps (which removed a couple of dependencies)

and then I rebuilt.


MikeCJohnsonUk (approved):
I'm presuming package-lock.json is a generated file? So no need to review directly?

Jim3Things:
Yes, package-lock.json is generated.  I'm not completely sure that we need to check it in, but have been doing so out of caution.

PR #286, opened 2021-05-16
Migrate observer UI to use redux

### LINKED ISSUE #281: Move observer to use redux

Jim3Things:
closes #281 

Note that there are several new libraries that this depends on that will need to be installed....

### DIFF
--- a/clients/observer/src/proxies/Session.tsx
+++ b/clients/observer/src/proxies/Session.tsx
@@ -4,14 +4,12 @@ import {UserPublic} from "../pkg/protos/admin/users"
 
 export interface SessionUser {
     name: string
-    password: string
     details: UserPublic
 }
 
-export function CreateSessionUser(val: any, name: string, password: string): SessionUser {
-    return  {
+export function CreateSessionUser(val: any, name: string): SessionUser {
+    return {
         name: name,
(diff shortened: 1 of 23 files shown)


clients/observer/src/proxies/Session.tsx
```diff
@@ -2,19 +2,18 @@
...
+}
 
-    constructor(val: any, name: string, password: string) {
-        this.name = name
-        this.password = password
-        this.details = UserPublic.fromJSON(val)
+export function CreateSessionUser(val: any, name: string): SessionUser {
+    return {
+        name: name,
+        details: UserPublic.fromJSON(val)
```
MikeCJohnsonUk, line 13: Lost the password? Or is it no longer required?

I guess I'm not sure how the logon logic works. Seeing that logon() still uses the password but nothing else seems to (where it did before?) I just couldn't quite join the dots.

Jim3Things: >It isn't needed anymore.  It was previously held after logon for the change password support - but that has now been removed from the observer UI.  So, we don't hold onto the password after the logon completes...

MikeCJohnsonUk (approved):

PR #288, opened 2021-05-20
Add Impact statements to inventoryd operations

Jim3Things:
Impact claims added to inventory state machine operations
Impact is moved to the span
uplift from child spans is supported in the text output
reason and span name updates are now done via setting attributes to the span
tsx form of the log entries are updated.  ts_ref is updated from generated code to track the change.

### DIFF
--- a/pkg/protos/log/entry.proto
+++ b/pkg/protos/log/entry.proto
@@ -47,7 +47,7 @@ enum Action {
     // relationships are strong - they can safely assume that both spans will
     // complete, they will execute in the same process, and that completion of the
     // overall trace ID is not complete until both are complete.
-    SpanStart = 3;
+    SpanStart = 1;
 
     // AddLink is used to place a the request point that may result in a linked
     // span.  It has an associated ID that is assigned by the active span at the
@@ -59,25 +59,22 @@ enum Action {
     // linked information as soft (optional) parent/child relationships.  If they
     // can put them into a logical execution tree, they do so.  If they cannot,
     // then they do not.
-    AddLink = 4;
+    AddLink = 2;
 
     // AddImpact is used to add an impact target to the span information.  The
     // impact value is a string stored in the text field, and is expected to
     // match the format used by the normal span KV structure (e.g. R:foo to 
     // indicate a read impact on component 'foo').
-    AddImpact = 5;
+    AddImpact = 3;
 }
 
 enum Severity {
     Debug = 0;
 
-    // This is the severity use to denote a purely explanatory entry
-    Reason = 1;
-
-    Info = 2;
-    Warning = 3;
-    Error = 4;
-    Fatal = 5;
+    Info = 1;
+    Warning = 2;
+    Error = 3;
+    Fatal = 4;
 }
 
 // Describe an impacted module
--- a/pkg/protos/log/entry.proto
+++ b/pkg/protos/log/entry.proto
@@ -40,14 +40,6 @@ enum Action {
     // child trace event.
     Trace = 0;
 
-    // UpdateSpanName and UpdateReason are directives to edit the containing span
-    // information.  The first replaces the span's name field, and the second
-    // replaces the span's reason text. This allows for better descriptions for a
-    // span once the details have been better understood - e.g. 'logging in a user'
-    // vs. 'logging in user "admin"'.
-    UpdateSpanName = 1;
-    UpdateReason = 2;
-
     // SpanStart is used to place the child span in the correct spot in the
     // sequence of events in the containing span.  It identifies the child span's
     // ID.  Structured formatters will expand the child span at this point in the
@@ -68,6 +60,12 @@ enum Action {
     // can put them into a logical execution tree, they do so.  If they cannot,
     // then they do not.
     AddLink = 4;
+
+    // AddImpact is used to add an impact target to the span information.  The
+    // impact value is a string stored in the text field, and is expected to
+    // match the format used by the normal span KV structure (e.g. R:foo to 
+    // indicate a read impact on component 'foo').
+    AddImpact = 5;
 }
 
 enum Severity {
@@ -105,20 +103,17 @@ message Event {
     // Formatted caller's stack trace
     string stack_trace = 5;
 
-    // The set of modules impacted, and the type of impact.
-    repeated Module impacted = 6;
-
     // Action to take when this trace is encountered.
-    Action event_action = 7;
+    Action event_action = 6;
 
     // Child's span ID.  Ignored if the action is not SpanStart.
-    string span_id = 8;
+    string span_id = 7;
 
     // Outgoing link ID.  Ignored if the action is not AddLink.
-    string link_id = 9;
+    string link_id = 8;
 
     // Real-world time when this event occurred.
-    google.protobuf.Timestamp at = 10;
+    google.protobuf.Timestamp at = 9;
 }
 
 // Describe a full correlated span, consisting of zero or more events.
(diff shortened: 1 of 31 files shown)


simulation/internal/services/inventory/tor.go
```diff
@@ -134,8 +137,10 @@ func torGetStatus(ctx context.Context, machine *sm.SM, msg sm.Envelope) bool {
 	occursAt := common.TickFromContext(ctx)
 
 	if m.Target.IsTor() {
+		tracing.AddImpact(ctx, tracing.ImpactRead, machine.Name)
```
MikeCJohnsonUk, line 140: When looking at this I was trying to figure out where machine.name came from. It seems the tor cable connection target is still being faked? The tor ports should now have this information in the Definition_Tor structs so it should be possible to replace fixConnection() with something derived from the stored inventory definitions. I'm assuming this would be a candidate for a future change.

simulation/internal/tracing/spanEx.go
```diff
@@ -111,6 +144,14 @@ func WithLink(sc trace.SpanContext, tag string) StartSpanOption {
...
 }
 
+// WithImpact states that the activity covered in the calling trace event had
+// the specified impact on the specified element.
+func WithImpact(impact string, element string) StartSpanOption {
+	return func(cfg *startSpanConfig) {
+		cfg.addImpact(ImpactKey, impact+":"+element)
+	}
+}
+
```
MikeCJohnsonUk, line 154: I presume if there are multiple impacts, there can be multiple WithImpact() calls in the options list on a trace call?

simulation/internal/services/inventory/inventory.go
```diff
@@ -155,20 +155,27 @@ func (s *server) initializeRacks() error {
...
 		}
 
-		s.racks[name] = newRack(ctx, name, r, s.timers)
+		s.racks[rack.Key] = newRack(
+			ctx,
+			rack.Key,
+			r,
+			rack.KeyIndexPdu,
+			rack.KeyIndexTor,
+			rack.KeyIndexBlade,
```
MikeCJohnsonUk, line 174: I'm not at all sure these should be public. They represent the etcd key(s) used for wildcard lookups of the class of objects represented in the name.

It maybe that this is code which needs to migrate to the inventory layer for the persisted state rather than the inventory service (do we need better naming here?)?

pkg/protos/log/entry.proto
```diff
@@ -40,14 +40,6 @@ enum Action {
...
     Trace = 0;
 
-    // UpdateSpanName and UpdateReason are directives to edit the containing span
-    // information.  The first replaces the span's name field, and the second
-    // replaces the span's reason text. This allows for better descriptions for a
-    // span once the details have been better understood - e.g. 'logging in a user'
-    // vs. 'logging in user "admin"'.
-    UpdateSpanName = 1;
-    UpdateReason = 2;
-
```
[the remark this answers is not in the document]
Jim3Things: That's correct - the enum covers the action values for trace events.  I found that there is a call that can be made to an open span that sets or replaces a specific attribute.  Since both of these are 'last one wins' settings, and they are not syncrhonized with the specific event in the span, I moved them from being trace events to just setting the span attribute.  That meant that these action values are no longer needed.

pkg/protos/log/entry.tsx
```diff
@@ -102,18 +100,15 @@ export function actionFromJSON(object: any): Action {
...
       return Action.SpanStart;
     case 4:
     case "AddLink":
       return Action.AddLink;
+    case 5:
+    case "AddImpact":
+      return Action.AddImpact;
     case -1:
     case "UNRECOGNIZED":
     default:
```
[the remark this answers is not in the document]
Jim3Things: I think you're asking about reusing the enum values, rather than leaving the old values as unused holes.  The rules say that we shouldn't, but in this case we probably could - because the value is never stored on disk and all bits are distributed at the same time.  But in general we shouldn't.  I can remove the hole in the numbering, if you'd prefer.

simulation/internal/services/inventory/blade.go
```diff
@@ -61,31 +61,34 @@ type blade struct {
...
 		holder:           r,
 		id:               id,
 		sm:               nil,
 		capacity:         messages.NewCapacity(),
-		architecture:     def.Arch,
+		architecture:     capacity.Arch,
 		used:             messages.NewCapacity(),
 		workloads:        make(map[string]*workload),
-		bootOnPower:      true,
+		bootOnPower:      def.BootOnPowerOn,
```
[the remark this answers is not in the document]
Jim3Things: Yeah, we should probably consistently do one or the other.  I can change this to GetXxx...


For Go, I'm also unconvinced that this does much of anything.  Java, at least, has the feature that GetXxx returns a value that covers missing, HasXxx returns true if it is present, and false if it is missing, and Xxx returns the raw value (including null).  Go doesn't have HasXxx, and GetXxx and Xxx seem to return the same things for non-object-reference values, from what I can tell...

simulation/internal/services/inventory/tor.go
Jim3Things: machine.Name is the state machine name value, which is set on creation.  The value currently used for hte Tor is rack.KiyIndexTor+tor-index (zero, currently).  This is the name that we discussed replacing with a simple moniker.


The inventoryd support currently hasn't incorporated the full definition information yet.  Coming soon.

simulation/internal/tracing/spanEx.go
Jim3Things: That's an excellent question.  I just looked, and they won't.  The code was carried over from the trace event version, and can now be simplified, as the span attribute handling is always 'last one wins'....

simulation/internal/services/inventory/inventory.go
Jim3Things: This should be replaced with something where I can get some opaque persisted moniker that I can send.  I'm currently using these as prefixes for constructing something that I can use as a moniker.

pkg/protos/log/entry.proto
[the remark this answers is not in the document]
MikeCJohnsonUk: Ok

pkg/protos/log/entry.tsx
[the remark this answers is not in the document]
MikeCJohnsonUk: I was simply asking about the symbolisation. I've no problem with the "holes"...

Ok

simulation/internal/services/inventory/blade.go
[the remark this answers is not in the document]
MikeCJohnsonUk: As far as I can see the only "useful" thing they auto generate is the missing check. Not sure it's always worth it however.

simulation/internal/clients/inventory/inventory.go
```diff
@@ -1162,124 +1157,225 @@ func (m *Inventory) writeInventoryDefinitionToStore(ctx context.Context, root *p
...
+}
 
-		storeRegion.SetDetails(region.GetDetails())
+// writeOneZone writes the records associated with the supplied zone to the store.
+func (m *Inventory) writeOneZone(
+	ctx context.Context,
+	regionName string,
+	zoneName string,
+	storeRegion *Region,
+	zone *pb.Definition_Zone) error {
```
MikeCJohnsonUk, line 1207: We don't need all of these arguments. In general, if we have the "cursor" object, we know the (string) names of all the parent components in the address. That is, if we have a *Region, we know the name of the region. Likewise, if we have a *Zone, we have the zone and region names.

Ok, just had the teleconf about the mechanical refactor. Given I don't see any functional issues, we can just mark this for later revision and move on.

MikeCJohnsonUk (approved):

PR #294, opened 2021-06-14
Supply details to UI, use defined TOR and PDU values

Jim3Things:
.. plus a few minor code cleanup items


simulation/internal/services/frontend/inventory.go
```diff
@@ -297,22 +299,49 @@ func transformRack(rd *pb.Definition_Rack) (*pb.External_Rack, error) {
...
+			t.Ports[k] = &pb.External_Tor_Port{
+				Port: port,
+				Observed: &pb.External_ObservedCable{
+					At:        0,
+					SmState:   fakeCableState(),
+					EnteredAt: 0,
+				}}
+		}
+
+		rack.Tors[i] = t
```
MikeCJohnsonUk, line 323: I think what we need here is a tor.GetExternal() and port.GetExternal() methods, or something like that, for the tor object rather than construct the External_Tor in line. I already have some similar methods elsewhere so nothing new. That would clean this path up a little. 

This would probably be better addressed in a separate change. I've added issue #295

simulation/internal/services/frontend/inventory.go
```diff
@@ -297,22 +299,49 @@ func transformRack(rd *pb.Definition_Rack) (*pb.External_Rack, error) {
...
+				Observed: &pb.External_ObservedCable{
+					At:        0,
+					SmState:   fakeCableState(),
+					EnteredAt: 0,
+				},
+			}
+		}
+
+		rack.Pdus[i] = p
 	}
```
MikeCJohnsonUk, line 346: Same sort of things for pdu.GetExternal()

Ref issue #295

simulation/internal/services/frontend/inventory.go
Jim3Things: agreed

simulation/internal/services/frontend/inventory.go
Jim3Things: agreed.

PR #304, opened 2021-06-28
Simulated inventory startup

Jim3Things:
Inventory now mocks the repair actions to power, connect, and boot the blades.
Network, power, and boot operations all incur delays.
Delay ranges are part of cloudchamber.yaml
The Delay request no longer has 'jitter' - any jitter is decided beforehand by the caller.
TimerExpiry messages now deliver the expiry message body rather than the TimerExpiry envelope, if a body was specified at creation.


simulation/internal/common/Range.go
```diff
@@ -0,0 +1,28 @@
...
+func (dr Range) String() string {
+	return fmt.Sprintf("%d:%d", dr.Low, dr.High)
+}
+
+// Pick returns a value in the acceptable range.
+func (dr Range) Pick() int64 {
+	delta := dr.High - dr.Low
+	if delta > 0 {
+		delta = rand.Int63n(delta + 1)
+	}
```
MikeCJohnsonUk, line 25: ... and if delta is negative? Yeah, it would be an illegal range, but still. The result would then always be dr.High?

simulation/internal/services/inventory/messages/TimerExpiry.go
```diff
@@ -41,18 +41,24 @@ func NewTimerExpiry(
...
-	if m.Target.IsTor() {
-		return r.ToTor(ctx, id, m)
+	if t.IsTor() {
+		return r.ToTor(ctx, id, msg)
 	}
 
-	if m.Target.IsBlade() {
-		return r.ToBlade(ctx, id, m)
+	if t.IsBlade() {
+		return r.ToBlade(ctx, id, msg)
```
MikeCJohnsonUk, line 61: Being picky, isn't this a switch statement? Or a candidate at least.

simulation/internal/config/settings.go
```diff
@@ -225,6 +244,22 @@ func newGlobalConfig() *GlobalConfig {
...
+				SetConnection: common.Range{
+					Low:  inventoryNetworkDelayLow,
+					High: inventoryNetworkDelayHigh,
+				},
+				Booting: common.Range{
+					Low:  inventoryBladeBootingDelayLow,
+					High: inventoryBladeBootingDelayHigh,
+				},
+			},
+		},
```
MikeCJohnsonUk, line 262: Somewhere in ReadGlobalConfig(), there probably needs to be a validation that the specified range is legal. I think this is the first 'composite' value we have.

For that matter, there are a couple of other data items that would benefit from some validation, e.g. port values should be less than 65536, Should there be "value is sane" type checks and so on? I feel an issue coming on?

simulation/internal/config/settings.go
```diff
@@ -168,6 +177,16 @@ type StoreType struct {
 	Test           StoreTypeTest
 }
 
+type DelayTypeInventory struct {
+	SetPower      common.Range
+	SetConnection common.Range
+	Booting       common.Range
+}
```
MikeCJohnsonUk, line 184: This does beg the question of whether on not these values should be part of the inventory property set? Or at least overrides on the underlying defaults supplied here. I can easily see, say, boot time varying from one blade type/configuration to another, or boot time being dependent on boot image, etc.

Actually, that has me wondering whether or not we have the inventory definitions quite right. Should we keep the inventory h/w specific things in in a catalog (such as power on delay), and then the current definition should have a reference to that blade type? Not sure how helpful that really is given that blades are inherently configurable and so exactly what would you separate out to put in such a catalog.

simulation/internal/common/Range.go
```diff
@@ -0,0 +1,28 @@
...
+	"fmt"
+	"math/rand"
+)
+
+// Range contains a pair of values that form a range of acceptable int64 values,
+// low to high, inclusive.
+type Range struct {
+	Low  int64
+	High int64
+}
```
MikeCJohnsonUk, line 13: I'm going to have to ask, should we use lower, upper or lower, length for our ranges? 

You can see the Azure storage influence showing up here I think...

Jim3Things: I thought about that, and thought that 'from x to y' was more understandable than 'from x for another y'.  YMMV.  We can always revisit, but I'd like to keep it this way for now.

simulation/internal/common/Range.go
Jim3Things: I've addressed this in the next rev.

simulation/internal/services/inventory/messages/TimerExpiry.go
Jim3Things: Good point.  I keep forgetting how general switch is.  Changed in the next rev.

simulation/internal/config/settings.go
Jim3Things: I've added validation for the global config in this next rev.

simulation/internal/config/settings.go
Jim3Things: I've been thinking about that as well.  Especially boot delays.  For now, let's keep them here.  We can always add them to blade definitions if we need to.

And yes, I'm pretty sure that we'll need to keep track of 'type' information that is shared between inventory instances.  It shouldn't be the catalog, but something derived from it.

simulation/internal/common/Range.go
MikeCJohnsonUk: I'm *very* happy keeping things this way. I kept on having trouble with "The Azure Way" (tm).

simulation/internal/services/inventory/testdata/inventory.yaml
```diff
@@ -0,0 +1,2057 @@
...
+                    Item: 
+                      Type: blade
+                      Id: 7
+                      Port: 0
+                  - Index: 8
+                    Wired: true
+                    Item: 
+                      Type: blade
+                      Id: 8
+                      Port: 0
```
MikeCJohnsonUk, line 212: You might wish to consider the use of TorTypes, (and for that matter PduTypes) from the file

wsimulation\internal\clients\inventory\testdata\Reference\inventory.yaml

which would allow you to collapse all the Tor (and Pdu) clauses to something like

    Tors:
      - Index: 0
        <<: *torType1

which would shorten the file considerably. Until we have the macro language though, there's not much to be done with blades (Parameter node value is the problem).

This is NOT required though as the current file will work just fine. Your call.

MikeCJohnsonUk (approved):
Just an observation on how inventory.yaml could be shortened, but the current version is sufficient. It can be left as is and optionally updated in a future change. Very much your call.

simulation/internal/config/settings.go
MikeCJohnsonUk: I didn't mean "The" catalog, just a catalog.

I'm ok with the rest.

simulation/internal/config/settings.go
MikeCJohnsonUk: Be careful what you wish for... ;-)

simulation/internal/common/Range.go
MikeCJohnsonUk: Ok.

simulation/internal/services/inventory/messages/TimerExpiry.go
MikeCJohnsonUk: Ok.

simulation/internal/services/inventory/testdata/inventory.yaml
Jim3Things: I'll leave it for now, but may come back to it soon.

PR #307, opened 2021-07-19
The Jet lag checkin

Jim3Things:
This is pretty mechanical.  I ran a code analysis and fixed a number of trivial errors - nothing deep.  I also updated npm, in which I found that the package-lock file had been missed earlier.  That is now included.  Finally, I removed a few known defunct files.

There is a lot of cleanup to do on the readme's, but that is for later.


simulation/docs/BuildEnvironmentSetupForWindows.md
```diff
@@ -1,4 +1,4 @@
-The following instructions have proven useful to setup a working
+The following instructions have proven useful to set up a working
```
MikeCJohnsonUk, line 1: We should be careful here. This is a generated file from a Word docx file which we earlier decided to not checkin. Do we want to re-visit that decision?

simulation/internal/README.md
```diff
@@ -1,13 +1,13 @@
...
 external packages into CloudChamber.
 
 # Services
 
-Most of the logic is embedded in microservices.  Each microservice is in a directory under 'services'.  All interactions
+The microservices contain most of the logic.  Each microservice is in a directory under 'services'.  All interactions
 with any microservice by the rest of CloudChamber is through gRPC.  External interactions are through RESTful patterns on http.
 
-Do be aware that most microservices use the store both as long term state, and as notification for actions.  
+Do be aware that most microservices use the store both as long term state, and as notification for actions.
```
MikeCJohnsonUk, line 10: Should there be a a "preserved" in "long term state". Hmm, not sure. 

Perhaps something like

Do be aware that most microservices use the store both for holding long term state, and as notification mechanism [or "source"?] for actions."

simulation/internal/clients/inventory/definition_test.go
```diff
@@ -72,7 +71,7 @@ func (ts *definitionTestSuite) stdRackDetails(suffix string) *pb.RackDetails {
 	}
 }
 
-func (ts *definitionTestSuite) stdPduDetails(ID int64) *pb.PduDetails {
+func (ts *definitionTestSuite) stdPduDetails(_ int64) *pb.PduDetails {
```
MikeCJohnsonUk, line 74: If the ID is truly not needed, shouldn't it just be removed completely?

simulation/internal/clients/inventory/definition_test.go
```diff
@@ -220,7 +219,7 @@ func (ts *definitionTestSuite) verifyStandardInventoryPdu(index int64, pdu Pdu)
 	}
 }
 
-func (ts *definitionTestSuite) verifyStandardInventoryTor(index int64, tor *Tor) {
+func (ts *definitionTestSuite) verifyStandardInventoryTor(_ int64, tor *Tor) {
```
MikeCJohnsonUk, line 222: If the ID is truly not needed, shouldn't it just be removed completely?

Applies to Pdu and Blade verification too.

simulation/internal/clients/store/store.go
```diff
@@ -179,7 +179,7 @@ func setDefaultNamespaceSuffix(suffix string) {
 	storeRoot.DefaultNamespaceSuffix = suffix
 }
 
-func (store *Store) connected(ctx context.Context) error {
+func (store *Store) connected(_ context.Context) error {
```
MikeCJohnsonUk, line 182: Any chance we can exclude store.go, store_test.go, storeapi.go, storeapi_test.go, DBInventory.go and errors.go from this checkin? I'm in the middle of changes to add the WithRevision() option on the ReadXxx() calls into the store.

simulation/internal/clients/store/store.go
```diff
@@ -188,7 +188,7 @@ func (store *Store) connected(ctx context.Context) error {
 	return nil
 }
 
-func (store *Store) disconnected(ctx context.Context) error {
+func (store *Store) disconnected(_ context.Context) error {
```
MikeCJohnsonUk, line 191: I've seen this style of change a number of times now. I presume this is the auto-formatter? If so, can we choose to just remove the arg if it's not being used? I completely fail to see the point of having to supply an arg which is definitively not used (unless it's necessary to satisfy an interface of course).

simulation/internal/clients/store/store.go
```diff
@@ -287,24 +287,25 @@ func cleanNamespace(ctx context.Context, testNamespace string) error {
...
-		return err
-	}
+		store.Disconnect()
 
-	store.Disconnect()
+		return nil
+	}
 
+	tracing.Fatal(ctx, "unable to allocate store context for pre-cleanup")
 	return nil
```
MikeCJohnsonUk, line 309: Ah, what? There's nothing wrong with the original. The tracing.Fatal() doesn't return. If the pretty printer can't handle that then a far better change would be to return an err on detecting the store was nil.

simulation/internal/services/frontend/httpErrors.go
```diff
@@ -219,28 +219,28 @@ func NewErrRackNotFound(name string) *HTTPError {
 
 // NewErrPduNotFound indicates that the rack was found but no pdu was found
 //
-func NewErrPduNotFound(rackid string, pduID int64) *HTTPError {
+func NewErrPduNotFound(rack string, pdu int64) *HTTPError {
```
MikeCJohnsonUk, line 222: I'm afraid I disagree with this change. The use of "ID" was deliberate since in the calling context rack and/or pdu will almost certainly be referring to the objects themselves, whereas here we just want the name/ID identifier and I'd suggest having the field names being the same as the object variable names is going to lead to confusion..

Now, if we were to use rackID, or likely better rackName, I'd be down with that.

Also applies to the other changes below.

simulation/internal/services/frontend/workloads.go
```diff
@@ -19,12 +19,12 @@ import (
...
 type Workload struct {
-	Name            string
-	TimeCreation    time.Time
-	TimeLasteUpdate time.Time
-	TimeLastStep    time.Time
-	StepCount       uint64
-	Enabled         bool
+	Name           string
+	TimeCreation   time.Time
+	TimeLastUpdate time.Time
```
MikeCJohnsonUk, line 24: It was the Ye Olde English version of 'Last' - ;-)

simulation/internal/tracing/spanEx.go
```diff
@@ -140,7 +140,7 @@ func WithLink(sc trace.SpanContext, tag string) StartSpanOption {
 // the specified impact on the specified element.
 func WithImpact(impact string, element string) StartSpanOption {
 	return func(cfg *startSpanConfig) {
-		cfg.impact = impact+":"+element
+		cfg.impact = impact + ":" + element
```
MikeCJohnsonUk, line 143: Gotta ask - why do these instances of a '+' operator get separating spaces added whereas an earlier instance had them removed.

Yeah, I know, reasons.

simulation/pkg/errors/errors.go
```diff
@@ -9,19 +9,6 @@ var (
...
-
-	// ErrStoreUnableToCreateClient indicates that it is not currently possible
-	// to create a client.
-	//
-	ErrStoreUnableToCreateClient = errors.New("CloudChamber: unable to create a new client")
-
-	// ErrUserUnableToCreate indicates the specified user account cannot be
-	// created at this time
-	ErrUserUnableToCreate = errors.New("CloudChamber: unable to create a user account at this time")
-
```
MikeCJohnsonUk, line 24: Are we sure these are not going to be required in the future?

simulation/pkg/protos/README.md
```diff
@@ -1,6 +1,6 @@
 This subtree contains the various proto definitions used by Cloud Chamber.
 
 Only the proto files are checked in.  The generated files are excluded from
-git.  If you make a change to a proto file you must fully build CloudChamber
+git.  If you make a change to a proto file, you must fully build CloudChamber
```
MikeCJohnsonUk, line 4: Ah, I think the comma is not required here?

Jim3Things:
Decided to bring this back as multiple smaller changes, and to actually clean the readme content as we do.

PR #309, opened 2021-07-21
 Add test to verify transaction read properly returns appropriate revision values on the individual records

MikeCJohnsonUk:
Add a test to verify that when multiple disjoint updates have occurred against a set of records being returned in a single transactional read request, all the records have the correct update revisions associated with the key, value pair.


simulation/internal/clients/store/store.go
```diff
@@ -1346,10 +1346,6 @@ func (store *Store) readTxn(ctx context.Context, request *Request) (response *Re
 	//
 	resp.Revision = txnResponse.Header.GetRevision()
 
-	// for _, r := range resp.Records {
-	// 	r.Revision = resp.Revision
-	// }
-
```
MikeCJohnsonUk, line 1352: This appears to be the result of some sort of merge wierdness resulting from my previous change. These lines are not supposed to be here.

simulation/internal/clients/store/store_test.go
```diff
@@ -684,6 +684,83 @@ func (ts *storeTestSuite) TestStoreWriteReadMultipleTxnPartial() {
...
+	assert.Equal(2, len(writeRequestComplete.Records))
+	assert.Equal(len(writeRequestComplete.Records), len(readRequest.Records))
+
+	// Attempt to read from the set of keys before they exist. As this is a "ConditionUnconditional" read
+	// request, it should succeed and produce an empty response.
+	//
+	readResponse, err := ts.store.readTxn(ctx, readRequest)
+	require.NoError(err, "Failed to read from store - error: %v", err)
+	require.NotNil(readResponse, "Failed to get a response as expected - error: %v", err)
+	assert.Equal(0, len(readResponse.Records), "Found %d records when none were expected", len(readResponse.Records))
```
Jim3Things, line 713: Nit, but you can use assert.Empty(readResponse.Records, ...)

simulation/internal/clients/store/store_test.go
```diff
@@ -684,6 +684,83 @@ func (ts *storeTestSuite) TestStoreWriteReadMultipleTxnPartial() {
...
+
+	testName := "TestStoreWriteReadMultipleTxnInversePartial"
+	key := testGenerateKeyFromName(testName)
+
+	writeRequestPartial := testGenerateRequestForWrite(1, key)
+	writeRequestComplete := testGenerateRequestForWrite(2, key)
+
+	readRequest := testGenerateRequestForReadWithCondition(len(writeRequestComplete.Records), key, ConditionUnconditional)
+
+	assert.Equal(1, len(writeRequestPartial.Records))
```
Jim3Things, line 703: Nit, but this can be assert.Len(writeRequestPartial.Records, 1).  Note that this function has the expected length as the second argument, which I don't much like...

Jim3Things (commented):
I'm fine with this - if you do make the nit changes I don't need to review again.  And I'm ok if you don't make them, also.

simulation/internal/clients/store/store_test.go
MikeCJohnsonUk: I like this so I'll use it.

simulation/internal/clients/store/store_test.go
MikeCJohnsonUk: That "args being the wrong way around" nixes it for me. This one I'll leave as is.

MikeCJohnsonUk:
Checked in after making use of one of the "nit" suggestions.

PR #312, opened 2021-07-31
Update Timer feature

### LINKED ISSUE #263: Combine timer functionality into listener & ticker

Jim3Things:
As preparation for saving the actual and observed states, this addresses two outstanding issues with the Timer feature.  First, it handle negative delays and already passed expiration times.  Second, the duplication between the Timer feature and the Timestamp feature are removed, with Timer becoming a pair of functions on the timestamp handling.

closes #263


simulation/internal/clients/timestamp/timer_test.go
```diff
@@ -18,42 +18,21 @@ type doneMsg struct {
...
 }
 
-func (ts *timerTestSuite) TestSimple() {
-	require := ts.Require()
-	assert := ts.Assert()
-
-	timers := NewTimers(ts.ep, ts.dialOpts...)
-
-	require.NotNil(timers)
-	ts.validateTimerState(timers, 0, 0, 1, false, 1)
```
MikeCJohnsonUk, line 28: What happened to validateTimeState()? No longer required? No longer appropriate?

Jim3Things: There is no timer state anymore (it's just a client call to the timestamp infra), so, yeah, no longer appropriate...

MikeCJohnsonUk (approved):

simulation/internal/clients/timestamp/timer_test.go
MikeCJohnsonUk: Ok.

PR #314, opened 2021-08-05
Makefile improvements

### LINKED ISSUE #302: Include cmdlet build in the overall makefile

Jim3Things:
- Filter out unit test support files from the production files list
- Add powershell cmdlet build into the general makefile.
- Add script_test command to cover tests that need a running cloudchamber service to execute.

closes #302


Makefile
```diff
@@ -205,13 +210,15 @@ SRC_ARTIFACTS = \
...
 
-
 ARTIFACTS = $(addprefix $(KIT_BUILD)/, $(notdir $(SRC_ARTIFACTS)))
 
 OBSERVER = \
 	$(OBSERVER_UI_BUILD)/asset-manifest.json \
 	$(OBSERVER_UI_BUILD)/index.html
 
+CMDLETS_OUT = \
+	$(CMDLETS_BUILD)/Debug/netstandard2.0/CloudChamber.Cmdlets.dll
```
MikeCJohnsonUk, line 220: Should this really be "Debug" in the path?

clients/powershell/Makefile
```diff
@@ -0,0 +1,28 @@
+CP = cp
+CP-RECURSIVE = $(CP) -r
+
+MD = mkdir -p
+RM-RECURSIVE = $(RM) -r
+
+TOUCH = touch --no-create
+
```
MikeCJohnsonUk, line 8: I'm beginning to think we are accumulating enough boilerplate that it'll be worth having a standard include file. What do you think? 

Not as part of this change, but I'd be interested in opinions.

Makefile
Jim3Things: It's what we're building at the moment.  When we've figured out how to do the packaging correctly, then I'm assuming that we'll cut over to a Release build.  I've just been holding off on going to a release build before then.

clients/powershell/Makefile
Jim3Things: Yeah, I agree.  I was thinking the same thing.

Makefile
MikeCJohnsonUk: Ok.

clients/powershell/Makefile
MikeCJohnsonUk: I've added issue #315

MikeCJohnsonUk (approved):

PR #317, opened 2021-08-09
Add macro expansion features to inventory definition yaml files

MikeCJohnsonUk:
A problem for inventory.yaml files is a large amount of repetition resulting from certain fields in nodes towards the leaf of the field tree needing to be unique.

This changes adds the ability to have some macro-ization of these fields leading to dramatic reductions of file sizes when used in conjunction with with yaml "type".

All the currently available macros have automatically generated value derived from previously defined fields (eg REGIONNAME, ZONENAME, RACKNAME, PDUID, TORID, BLADEID), fixed expansions (eg NODENAME) or special actions which generate values on demand (eg GUID and UUID).

These string values are substituted for the matching $<macro>$ as the inventory definition is loaded into the system.


simulation/internal/clients/inventory/reader.go
```diff
@@ -190,6 +192,161 @@ type bladeExBootInfo struct {
...
 
+
+
+// Macro replacement
+// Convert all to using an Index as the key, then have NAME as a separate field. Needs to be a top level node in the yaml to allow easy replacement on node merge when aliased.
+
+// GENERATED values
+// REGIONID is the index value of the nearest enclosing region
+// ZONEID is the index value of the nearest enclosing zone
+// RACKND is the index value of the nearest enclosing rack
```
MikeCJohnsonUk, line 203: Yes, I just noticed type where RACKND should be RACKID

simulation/internal/clients/inventory/reader.go
```diff
@@ -190,6 +192,161 @@ type bladeExBootInfo struct {
...
+
+	return sm
+}
+
+func (sm *symbolMap) set(sym string, val string) string {
+	sm.symbols[sym] = val
+	return val
+}
+
+func (sm *symbolMap) setFronInt64(sym string, val int64) string {
```
MikeCJohnsonUk, line 265: Typo in function name - should be setFromInt64()

simulation/internal/clients/inventory/reader.go
```diff
@@ -190,6 +192,161 @@ type bladeExBootInfo struct {
...
+			str = strings.ReplaceAll(str, fmt.Sprintf("$%s$", k), val)
+		}
+	}
+
+ 	return str
+}
+
+func (sm *symbolMap) setRegion(name string, id int64) (int64, string) {
+	if id == 0 {
+		id = atomic.AddInt64(&sm.nextRegionId, 1)
```
Jim3Things, line 300: I don't think you need to use an atomic increment here.  In Go, maps are thread-safe, so line 303, I think, means that sm must have a single thread accessing it anyway...

Jim3Things (approved):
The new format looks great.  One comment, you can fix or not, as per your wish.  I don't need to look at it again, so assume approval.

simulation/internal/clients/inventory/reader.go
MikeCJohnsonUk: I'll do some digging to reassure myself then remove this if warranted. I'll do that (potential) removal in a follow-up PR.

MikeCJohnsonUk:
I'll take care of the minor issues in a follow-up PR rather than trying to fixup this one. Expect to see that shortly.

PR #319, opened 2021-08-14
Add rack actual view support

Jim3Things:


simulation/internal/clients/inventory/mocks.go
```diff
@@ -14,50 +15,100 @@ type MockStore struct {
...
+		context.Background(),
+		tracing.WithName("Mock Store: Disconnect"))
+	span.End()
+
 	_ = x.Called()
 }
 
 func (x *MockStore) Update(ctx context.Context, r namespace.KeyRoot, n string, rev int64, v string) (revision int64, err error) {
+	_, span := tracing.StartSpan(
+		ctx,
```
MikeCJohnsonUk, line 38: One thought occurs, and this is more of a question. At some point, I expect to be adding some WithXxxx() options on some of the corresponding base calls. I presume, that means adding them here too? Even if that have no effect. Think WithRevision() and so on.

Jim3Things: If the signature to a function defined in the store.Operation interface changes, then it needs to change here as well.  I'm guessing from your example that this would be the case - so, if so, yeah.

MikeCJohnsonUk: Ok. That matches my expectations.

MikeCJohnsonUk (approved):

PR #325, opened 2021-08-31
Plumb Tor & Pdu actual view data to the UI

Jim3Things:


clients/observer/src/SimulatedInventory/PDU.tsx
```diff
@@ -57,6 +62,22 @@ function powerDetail(onOff: boolean) {
     }
 }
 
+function cableDetail(cable: Cable, pdu: External_Pdu) {
+    const port = pdu.ports.get(cable.portIndex)
+    const hwString = hardware_HwTypeToString(cable.target.type)
+    const wired = port?.port.wired ? "connected" : "disconnected"
```
MikeCJohnsonUk, line 68: For my curiosity's sake, what's the port?.port.wired syntax about?

simulation/internal/services/frontend/inventory.go
```diff
@@ -302,27 +302,41 @@ func transformRack(ctx context.Context, rd *pb.Definition_Rack, ra *pb.Actual_Ra
...
+			t.Actual = &pb.External_Tor_ActualState{
+				SmState:   ta.GetSmState(),
+				EnteredAt: 9,
+				Cables:    pb.CloneActualCables(ta.GetCables()),
+			}
+
+			t.Observed = &pb.External_Tor_ObservedState{
+				At:        common.TickFromContext(ctx),
+				SmState:   ta.GetSmState(),
+				EnteredAt: 9,
```
MikeCJohnsonUk, line 337: So, if I understand this correctly, we observe the entry to a state in the same clock tick ('9') as the entry to the actual state?

I must have that wrong. We shouldn't be entering any state at a fixed tick. I may need to change this comment.

clients/observer/src/SimulatedInventory/PDU.tsx
MikeCJohnsonUk: After the explanation, makes sense.

Ok.

simulation/internal/services/frontend/inventory.go
MikeCJohnsonUk: After the explanation, makes sense.

Ok.

MikeCJohnsonUk (approved):

PR #326, opened 2021-09-04
Integrate redux for inventory updates, first round of UI perf work

Jim3Things:
This change has two main parts: first, the inventory in the UI is now held in redux, and the updates to the display are driven from that.  This has eliminated the full resplat that we were seeing on an inventory re-load.

This also exposed that we've once again tipped over the edge on perf.  There are some changes here to help with perf, but there is still a lot of work to do.  Analysis shows that the log display is definitely hitting us, and I'm far from convinced that we're in the clear on the inventory display.  Expect more perf work coming.

Note that the perf work has forced some protocol incompatibilities.  I've opted to avoid the SxS proto changes just now in order to help the perf work along.


simulation/internal/services/frontend/inventory.go
```diff
@@ -78,8 +79,9 @@ func handlerRacksList(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	racks := make([]*pb.External_RackSummary, 0, memoData.RackCount)
```
MikeCJohnsonUk, line 82: May need to be a little careful here. I don't happen to recall whether the base dbInventory.GetMemoData() returns information per-region or per-zone (using default region/zone).

I think there is also a per-zone GetMemoData(), or at least there was?

simulation/pkg/protos/inventory/external.proto
```diff
@@ -55,6 +55,8 @@ message External {
             map<int64, Cable> cables = 3;
         }
         ActualState actual = 4;
+
+        int64 index = 5;
```
MikeCJohnsonUk, line 59: This corresponds to the Id field right?

simulation/internal/services/frontend/inventory.go
Jim3Things: There is, and I just checked - the dbInventory.GetMemoData does a GetDefaultZoneSummary, so I think this is ok.

simulation/pkg/protos/inventory/external.proto
Jim3Things: I think so?  If by id, you mean what had been the key in the in-rack map - so the id for the item in the rack, then yes.  There's no Id field in the proto files, so I wanted to check if I understood correctly.

simulation/internal/services/frontend/inventory.go
Jim3Things: fixed in the latest update.  thanks.

simulation/pkg/protos/inventory/external.proto
MikeCJohnsonUk: I'm referring to the Id that originates in the inventory.yaml file. Typically we've been using indices from 1 upwards, but there is no requirement that that be so. As long as the Id from the inventory file was used as a key to a map then no problem. as the Id was "preserved". If the Id is now present as a field within the object, then this should still work. I was just concerned that we might, possibly inadvertently, be using that value as the index to an array. For example, consider a rack with two pdus, one with Id of (say) 10, and the other with Id 20. Given that we size these arrays for the appropriate number of objects, in this case two, we clearly should not be using the "Id" as a index into that array.

Does that help explain my concern?

MikeCJohnsonUk: Actually, do we even need to retain these "Id" fields as numeric? Is there a benefit in them being anything other than a string? Just curious.

Jim3Things: Following up on an offline discussion: the index field is the value from the yaml field.  I've kept it as a number just for simplicity for now.  It turns out that we use it only for header display in the UI, so there's not much exposure to it at all.

MikeCJohnsonUk (approved):

PR #328, opened 2021-09-28
Normalize inventory redux data

Jim3Things:
This change normalizes the incoming inventory data into a set of redux tables.

The selectors now use the normal useAppSelector, and are 'correct' selectors.

No work on perf yet.  This is structural cleanup.


Jim3Things:
As per discussion around travel, I'm committing this and leaving the branch for later review.

MikeCJohnsonUk (commented):
Ok.

PR #329, opened 2021-09-29
Minor observer change

Jim3Things:
This makes three small changes:
- the old project name is removed from the last files that still had it.
- code inspector tools suggested an await was missing, which has been added.
- fake usage is now constructed more dynamically, and not loaded into a blade unless the blade is working.
- removed an obsolete source file


Jim3Things:
committing now, leaving branch for later review

MikeCJohnsonUk (commented):
Ok

PR #330, opened 2021-10-01
normalize the log trace data

Jim3Things:
This change normalizes the log trace structures into redux.

The organizer class is removed.

I've also pulled the 'max severity' icon on the spans, as that has a lot of computational load and I'm not sure how much help it gives.

The complex redux slices are split into their own modules.

Initial unit tests for the normalizeLog* functions added.


Jim3Things:
Committing and retaining for review, due to travel.

clients/observer/src/store/NormalizeLog.test.tsx
```diff
@@ -0,0 +1,132 @@
...
+        linkSpanID: "",
+        linkTraceID: "",
+        impacted: [],
+    })
+
+    const traceEntry3: Entry = new Entry({
+        name: "testSpan3",
+        spanID: "CH001",
+        parentID: "0002",
+        traceID: "000000004",
```
MikeCJohnsonUk, line 97: Is this supposed to have the same traceId value as traceEntry2?

MikeCJohnsonUk (commented):
Just a couple of probably minor questions...

clients/observer/src/store/NormalizeLog.test.tsx
Jim3Things: Yes.  The traceID is the tag that groups multiple spans together into a 'logical unit'.  We don't do much with it, but that seems to be the case.

clients/observer/src/store/NormalizeLog.tsx
```diff
@@ -0,0 +1,194 @@
...
+// arrays contain the keys to remove.
+export type NormalizedLogRemoval = {
+    removeEntries: string[],
+    removeEvents: string[],
+    removeLinks: string[],
+    removeRoots: string[],
+}
+
+function linkFQName(sourceSpan: string, sourceTrace: string, linkId: string): string {
+    if (linkId === "" || linkId === undefined) {
```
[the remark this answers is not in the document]
Jim3Things: If the test were to a field in linkId, then yes.  I believe that either order should be equivalent when there are no sub-fields involved.

clients/observer/src/store/NormalizeLog.test.tsx
MikeCJohnsonUk: Ok

clients/observer/src/store/NormalizeLog.tsx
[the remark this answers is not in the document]
MikeCJohnsonUk: Ok.

MikeCJohnsonUk:
Ok.

PR #331, opened 2021-10-02
Minor display improvement

Jim3Things:
Settings display now shows the trace retention limit - this change makes that line fit into the rest of the display better.


Jim3Things:
Committed, but leaving the branch for later review.

MikeCJohnsonUk (commented):
Ok.

PR #332, opened 2021-10-14
Simulate startup repair sequences, add faked workloads

Jim3Things:
This change has inventoryd store the observed view for a blade with a 1-tick delay.  It also adds workloads incrementally to the working blades, stopping when the blade gets full.

The fake observed view is removed from the frontend.  The fake usage creation is removed from the observer.

The rack observed state is minimal, just enough to get the blade observed state.
The faked initial repair sequence does _very_ minimal error handling, as it will be replaced by the actual repair manager.

Unit test for this are largely absent at this point, but will be coming for any stable portion in a future update.


simulation/internal/services/inventory/inventory.go
```diff
@@ -229,3 +268,173 @@ func (s *server) connectBladeAfterPower(r *Rack, b *blade, rsp chan *sm.Response
...
+		if !ok {
+			store = msg.b.storedBlade.Observed()
+			observers[key] = store
+		}
+
+		store.Observe(common.TickFromContext(ctx), msg.save)
+
+		_, err := store.Create(ctx)
+		if err != nil {
+			_, err = store.Update(ctx, true)
```
MikeCJohnsonUk, line 354: Is there something the Create() is not doing that requires a following Update(). Could these perhaps be merged? Or are you looking for a separate/specific event?

simulation/internal/services/inventory/inventory.go
```diff
@@ -229,3 +268,173 @@ func (s *server) connectBladeAfterPower(r *Rack, b *blade, rsp chan *sm.Response
...
+		}
+	}
+}
+
+// startWorkloads adds workload instances to a blade until the first workload
+// encounters an error, presumably due to resource exhaustion.
+//
+// It starts 1-3 instances per simulated time tick, then waits a tick, and
+// continues until the resource limits are hit.
+func (s *server) startWorkloads(r *Rack, target *messages.MessageTarget) {
```
MikeCJohnsonUk, line 377: I'm a little confused, is this a production routine, or does it belong to a test operation?

MikeCJohnsonUk (commented):
I don't think there was anything especially alarming.

PR #335, opened 2022-01-20
Move UI out of Typescript/Javascript, part 1

Jim3Things:
This is the first part of the change to remove all Typescript and NPM from the CloudChamber project.  The new Observer is written in C#, using WPF with limited dependencies.

There is much more work to do, but there's enough here to warrant an initial checkin.


Jim3Things:
Merging prior to review, based on team availability.  The branch will be retained for later review.

PR #336, opened 2022-01-29
Conversion to wpf, part 2

Jim3Things:
Get the inventory display and view models into a reasonable state.


Jim3Things:
Committing prior to review, based on team availability.  Will keep branch around for later review.

PR #340, opened 2022-02-06
Move login to use the Material dialog support.

Jim3Things:


Jim3Things:
Committing prior to review, based on team availability.  Branch retained for later review.

PR #344, opened 2022-02-18
Merge observerUI and cmdlets into general makefile and kitting.

Jim3Things:
.. and remove all typescript & npm dependencies.


Jim3Things:
I will also commit this prior to review and leave the branch around for later comment.  Hopefully we'll be able to return to review-before-commit soon...

PR #348, opened 2022-02-26
Move deployments directory to cloud chamber root

Jim3Things:
This is a small change that moves the output directory (deployments) from the simulation project to the top level.


Jim3Things:
commit before review, retaining branch

PR #350, opened 2022-03-03
Some trace cleanup, simplify observation pass, add Tor observed state

Jim3Things:


Jim3Things:
continuing commit before review, retaining branch for later review.

PR #351, opened 2022-03-05
Removal of opentelemetry dependency, step 1

Jim3Things:
OpenTelemetry has turned out to be much too much for our needs, and not quite the right fit for the rest.  So we'll be removing this dependency.  In order to simplify other changes going on, this first step temporarily moves the sources into the main cloudchamber project and out of the vendor tree.  It also removes some plugins that CloudChamber never uses.


Jim3Things:
Continuing commit before review, retaining branch

PR #355, opened 2022-03-20
Implement faster/slower run buttons

Jim3Things:
Technical debt: these buttons have been NYI until the activity log was cleaned up some.  Now implemented with a range of 1-5 ticks / second.


Jim3Things:
Commit before review, retaining branch for review.

MikeCJohnsonUk (commented):
Ok.

PR #357, opened 2022-03-21
Trace exporter infrastructure and stack trace removal

Jim3Things:
This is also some minor technical debt removal.  It removes the no-infra and no-stack options from the trace sink handling, as those filters are applied before the entries get to the trace sink.


MikeCJohnsonUk (approved):
Ok

PR #359, opened 2022-03-25
Remove obsolete vendor dependencies

Jim3Things:
These appear to no longer being used.  I've cleaned, built, run unit tests, and run the services successfully.


MikeCJohnsonUk (approved):
That was a tough one.

PR #362, opened 2022-06-12
Update two dependencies

Jim3Things:
Update viper & testify dependencies


MikeCJohnsonUk (approved):
That was a lot less dramatic than I'd feared.

PR #365, opened 2022-07-15
Updates to describe current tool chain and build requirements

MikeCJohnsonUk:
Update to indicate current tested version of elements of the toolchain
Now build is go mod based, use appropriate method of installing ProtoC language processors
Remove references to separate UI repository
Remove references to npm and related files.


simulation/docs/BuildEnvironmentSetupForWindows.md
```diff
@@ -672,7 +613,7 @@ and explain how to use an alternate location in a later section
     speed up the build. The effects will depend on the computer’s
     configuration.
 
-3. Start the Etcd store
+1. Start the Etcd store
```
Jim3Things, line 616: Why are these all 'step 1'?

Jim3Things (changes requested):

simulation/docs/BuildEnvironmentSetupForWindows.md
MikeCJohnsonUk: The actual number used in the MD source doesn't matter, it just needs to be a number. The renderer will ensure properly sequenced numbered sections regardless of the specific numbers in the source. So using '1' seemed as good as any and removed any potential confusion about specific numbers in the rendered doc being different from the MD source.

Jim3Things: ok.

PR #367, opened 2022-08-15
Minor cleanup and first use of generics

Jim3Things:
This change replaces a grpc obsolete function with the current equivalent, fixes some typos, and introduces the use of generics in two places: the type specific Max functions are replaced, and the PDU/TOR actual & observed view processing is greatly simplified by some common generic support functions.


Jim3Things:
Committing with review to follow

simulation/internal/common/math.go
```diff
@@ -1,16 +1,7 @@
...
-		return b
-	}
-
-	return a
-}
-
-// MaxInt is a helper function to return the maximum of two int values
-func MaxInt(a int, b int) int {
+// Max is a helper function to return the maximum of two values
+func Max[Number int | int64](a, b Number) Number {
```
MikeCJohnsonUk, line 4: should the type of Number perhaps be "constraints.Ordered" rather than just the set of int and int64? Or at least "~int | ~int64"?

MikeCJohnsonUk (commented):
I think I get most of this. Only one (minor) question which in no way requires any change.

PR #369, opened 2022-09-13
Further conversion to generics, update to Go 1.19

Jim3Things:
Most of this checkin is the result of a fresh go fmt on the project.  It has gotten even more opinionated than it was.  It won't say I like it, but go fmt is pretty much expected.  Better everyone's bad coding style than a good one, AFAICT.

The actual changes are for some remaining conversions to generics in the inventory client.


simulation/internal/clients/inventory/api.go
```diff
@@ -81,7 +89,6 @@ type ItemStoreOperations interface {
 
 // KeyProvider specifies the functions that an element that has a location in
 // the store must provide in order to support the addressing functions.
-//
```
MikeCJohnsonUk, line 84: You were probably expecting this, but for me, this is a big problem. Those comments were added specifically to make things readable when using non-colourizing editors. Removal introduces a readability issue. As such, I'm really not happy about this at all. 

If gofmt insists on doing this, I'd vote for ditching gofmt and it's unwanted opinions.

simulation/internal/clients/inventory/leafViewsOfT.go
```diff
@@ -0,0 +1,249 @@
...
+// SetState sets the stored state machine state identifier to the supplied
+// value.
+func (cls *CoreLeafState[T]) SetState(state T) {
+	cls.state = state
+	cls.resetRevision()
+}
+
+// Equal returns true if the supplied argument is logically equal to this one.
+func (cls *CoreLeafState[T]) Equal(x ActualAccess[T]) bool {
+	return x != nil &&
```
MikeCJohnsonUk, line 106: Should it be legal to allow x to be nil?

cf. other functions, eg Load() below, there is no explicit nil test...

Is this one of those convention things?

simulation/internal/clients/inventory/leafViewsOfT.go
```diff
@@ -0,0 +1,249 @@
...
+		x.GetTimes().LastGoodObservation,
+		x.GetTimes().FailedObservationDetails)
+}
+
+// Equal is used to provide a simple equality check to determine if the current
+// view matches the one supplied. Typically, it is used when looking to see if
+// the record has been changed.
+func (cov *CabledObservedView[T]) Equal(x CabledObservedAccess[T]) bool {
+	return x != nil &&
+		cov.CabledState.Equal(x) &&
```
MikeCJohnsonUk, line 231: Should we either commit to Equal() handling x being nil or not? I'd vote for not, but I could be convinced...

simulation/internal/clients/inventory/pdu.go
```diff
@@ -264,9 +257,7 @@ type PduViewActual struct {
 func newPduViewActual(p *Pdu) *PduViewActual {
 	act := &PduViewActual{
 		CabledActualView: CabledActualView[pb.PduState_SM]{
-			CabledState: CabledState[pb.PduState_SM]{
-				commonActualCabledView: newCommonActualCabledView(p.tables.Actual(), p),
-			},
+			CabledState: NewCabledState[pb.PduState_SM](p.tables.Actual(), p),
```
MikeCJohnsonUk, line 260: Much cleaner.

simulation/internal/services/frontend/db/users.go
```diff
@@ -177,10 +174,10 @@ func (m *Users) Read(ctx context.Context, name string) (*pb.User, int64, error)
 // expected (match) revision
 //
 // NOTE: Currently, this uses a *VERY* clumsy read/modify/write action. What is
-//       really needed here is to provide the ability to feed an action into the
-//       Update() routine itself to allow the caller to selectively update
-//       individual fields from within the transaction.
 //
```
MikeCJohnsonUk, line 177: Ah, another gofmt victim perhaps? Adding a blank comment in the middle of a paragraph?

simulation/internal/services/frontend/workloads.go
```diff
@@ -16,8 +16,8 @@ import (
 // Workload is a representation of a specific workload
 //
 // TODO This is just a placeholder until we have proper workload definitions
-//     held in a persisted store (Etcd)
 //
+//	held in a persisted store (Etcd)
```
MikeCJohnsonUk, line 20: Seriously?

MikeCJohnsonUk (changes requested):
I think the gofmt folks have lost the plot. They're removing blank comments to butt dissimilar constructs up against each other, and then start adding blank comments lines in the middle of sentences within a text block.

Can we please stop using this?

simulation/internal/clients/inventory/api.go
Jim3Things: I've reverted the files that weren't otherwise changed in the generics work.  That does leave some of the trailing blank comments removed.  I'm honestly not sure what to do about go fmt overall.

simulation/internal/clients/inventory/leafViewsOfT.go
Jim3Things: Probably the right check is to say that either x is not nil, or both are nil.  And load probably should set the values to defaults if the source is nil.  I'll update all with the next commit on this thread.

simulation/internal/clients/inventory/leafViewsOfT.go
Jim3Things: There's a common case where there is no prior state and we're trying to see if this state is the same.  Alllowing nil to be passed makes that easier at the calls site(s).

simulation/internal/services/frontend/db/users.go
Jim3Things: reverted.

simulation/internal/services/frontend/workloads.go
Jim3Things: reverted.

simulation/internal/clients/inventory/leafViewsOfT.go
Jim3Things: I've decided to leave load alone, but have fixed the equal calls.

simulation/internal/clients/store/TestSuiteCore.go
```diff
@@ -12,15 +12,15 @@ import (
 )
 
 var (
-	lock        sync.Mutex
-	initialized = false
+	lock         sync.Mutex
+	initialized  bool = false
```
MikeCJohnsonUk, line 16: Just a heads up: I'm happy with this either way, but it looks like a valid change that got reverted. Your call whether or not to accept the original modification.

simulation/internal/common/decode.go
```diff
@@ -18,12 +18,12 @@ func (e errInMatch) Error() string {
...
-//   - When there is no error, the boolean array describes which keywords were matched .
-//     An element in that array is true if the element with the same index in the keywords
-//     array was matched.
+// - It returns an error if no occurrence of the source string is found in the keyword array.
+// - When there is no error, the boolean array describes which keywords were matched .
+//  An element in that array is true if the element with the same index in the keywords
+//  array was matched.
 func Decode(keywords []string, source string) ([]bool, error) {
-	var test = strings.Split(source, ",")
+	var test []string = strings.Split(source, ",")
```
MikeCJohnsonUk, line 26: Another valid change that got reverted? I'm happy either way. Your call.

MikeCJohnsonUk (approved):

simulation/internal/clients/store/TestSuiteCore.go
Jim3Things: Yes, they were in a file I'd otherwise touched.  Since it is logically the same, I reverted down to the files I touched.  Same for the next one.

PR #370, opened 2022-09-19
Cleanup

Jim3Things:
I updated dependencies in the C# projects.

I also converted SM to use generics, and removed global access to most internal SM state.  I kept Name, as that is decoration and not used for any logic in SM itself.

Finally, I updated the obsolete use of ioutil.ReadAll


simulation/internal/services/inventory/blade_test.go
```diff
@@ -107,46 +107,45 @@ func (ts *BladeTestSuite) TestGetStatus() {
...
-	for _, b := range rackDef.blades {
-		b.EnsureAllCallsUsed()
-		b.actual.EnsureAllCallsUsed()
+	for _, mb := range rackDef.blades {
+		mb.EnsureAllCallsUsed()
+		mb.actual.EnsureAllCallsUsed()
 	}
 
-	b := rackDef.blades[0]
+	b := r.blades[0]
```
MikeCJohnsonUk, line 115: Changed to the mocked rack set rather than the dummy set used to construct the rack set? Now, without being sure of the difference, should we perhaps be using the mocked set throughout? Including for the checks above? It *looks* like this is (now?) correct but I'm uncertain.

simulation/internal/services/inventory/pdu_test.go
```diff
@@ -588,6 +574,7 @@ func (ts *PduTestSuite) TestStuckCablePduOff() {
 	ctx, r := ts.createAndStartRack(context.Background(), 2, true, true)
 
 	p := r.pdus[0]
+	mp := p.storedPdu.(*MockedPdu)
```
MikeCJohnsonUk, line 577: Is there perhaps a particular reason for moving this up here? Perhaps the call is no longer possible after the off state has been achieved? Just curious...

simulation/internal/services/inventory/blade_test.go
Jim3Things: so what happened is that this test is old enough that the naming was pretty inconsistent.  I've ended up with b for the standard blade reference, and mb for the reference to the underlying blade mock instance (i.e. reaching around behind the client interface).  I now needed the standard blade reference, so I fixed the naming.

simulation/internal/services/inventory/pdu_test.go
Jim3Things: No special reason - I had done that elsewhere and was just trying to make this similar.

MikeCJohnsonUk (approved):

